Issue #84, opened 2018-06-05
Return URL Rewrites for products


paliarush:
As a Magento developer, I want GraphQL to return enough data so I can construct category-inclusive product urls so I can render context-aware product links on category pages and support the built-in "use categories path for product URLs" feature. 

`products` query already supports [canonical URL](https://github.com/magento/graphql-ce/issues/2) for products, this URL is returned as if it was generated by search  with no category added to the URL. For example `https://magento.store/some-product.html`. 

Sometimes it makes sense to have category added as part of SEO friendly URL, for example `https://magento.store/some-category/some-product.html`. It can be useful while rendering links to products from the specific category page.

AC:
- Add `url_rewrites`: `![UrlRewrite]` field to `ProductInterface` GraphQL type
- `UrlRewrite` GraphQL type should have the following fields:
    - `url`: `String`
    - `parameters`: `[HttpQueryParameter]`. `HttpQueryParameter` type should have fields:
        - `name`: `String`
        - `value`: `String`
- It should be possible to reuse the same types and resolvers for other entities in the future, e.g. for categories or CMS pages
- Redirect rewrites must be excluded from the response (see rewrites grid in the admin panel for more info)
- Modularity should be taken into account and new dependencies between modules should not be introduced. `ProductInterface` schema should be modified via changes in `Magento/CatalogUrlRewriteGraphQl` module
- Changes are covered with web API functional tests. See [\Magento\GraphQl\Catalog\ProductViewTest::testQueryAllFieldsSimpleProduct](https://github.com/magento/magento2/blob/b3372064f574fd5aece0617009ee317db589febe/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/ProductViewTest.php#L35) as an example.

naydav:
https://github.com/magento/graphql-ce/pull/107

PR #107, opened 2018-06-30
Added URL rewrites data to the product interface

rogyar:
### Description
Added URL rewrites data to the GraphQL product interface. So, now we have a possibility to fetch all URL rewrites for a product with request URL data and target URL parameters.

Also, the old `UrlRewrite` resolver has been renamed to `EntityUrl` to make the class name match its purpose. 

1. https://github.com/magento/graphql-ce/issues/84

### Manual testing scenarios
- Create a product with name "configurable" and assign it to a category
- Use the following request to fetch the product data:

```
{
    products (search:"conf") {
        items {
            name,
            sku,
            description,
          	url_rewrites {
              url,
              parameters {
                name,
                value
              }
            }
        }
    }
}
```

- You should see 2 URL rewrites in the response like in the following example:

```
{
  "data": {
    "products": {
      "items": [
        {
          "name": "Configurable",
          "sku": "Configurable",
          "description": null,
          "url_rewrites": [
            {
              "url": "configurable.html",
              "parameters": [
                {
                  "name": "id",
                  "value": "5"
                }
              ]
            },
            {
              "url": "some-category/configurable.html",
              "parameters": [
                {
                  "name": "id",
                  "value": "5"
                },
                {
                  "name": "category",
                  "value": "3"
                }
              ]
            }
          ]
        }
      ]
    }
  }
}
```


app/code/Magento/UrlRewriteGraphQl/Model/Resolver/EntityUrl.php
```diff
@@ -0,0 +1,161 @@
...
+use Magento\Framework\GraphQl\Config\Element\Field;
+use Magento\Framework\GraphQl\Query\Resolver\Value;
+use Magento\Framework\GraphQl\Query\Resolver\ValueFactory;
+use Magento\Framework\GraphQl\Query\ResolverInterface;
+use Magento\Store\Model\StoreManagerInterface;
+use Magento\UrlRewrite\Model\UrlFinderInterface;
+use Magento\UrlRewriteGraphQl\Model\Resolver\UrlRewrite\CustomUrlLocatorInterface;
+
+/**
+ * UrlRewrite field resolver, used for GraphQL request processing.
```
vkublytskyi, line 19: As EntityUrl and UrlRewrite are pretty similar it is better to describe their purpose in class comments (now both classes have same description so to understand what is needed source code should be explored)

app/code/Magento/CatalogGraphQl/etc/schema.graphqls
```diff
@@ -280,6 +280,7 @@ interface ProductInterface @typeResolver(class: "Magento\\CatalogGraphQl\\Model\
     manufacturer: Int @doc(description: "A number representing the product's manufacturer")
     categories: [CategoryInterface] @doc(description: "The categories assigned to a product") @resolver(class: "Magento\\CatalogGraphQl\\Model\\Resolver\\Categories")
     canonical_url: String @doc(description: "Canonical URL") @resolver(class: "Magento\\CatalogGraphQl\\Model\\Resolver\\Product\\CanonicalUrl")
+    url_rewrites: [UrlRewrite] @doc(description: "URL rewrites list") @resolver(class: "Magento\\UrlRewriteGraphQl\\Model\\Resolver\\UrlRewrite")
```
naydav: Looks like we need to move all of this functionality to `CatalogUrlRewriteGraphQl`

app/code/Magento/UrlRewriteGraphQl/Model/Resolver/EntityUrl.php
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public function __construct(
+        UrlFinderInterface $urlFinder,
+        StoreManagerInterface $storeManager,
+        ValueFactory $valueFactory,
+        CustomUrlLocatorInterface $customUrlLocator
+    ) {
+        $this->urlFinder = $urlFinder;
+        $this->storeManager = $storeManager;
+        $this->valueFactory = $valueFactory;
```
naydav: Pls remove using Factory

naydav (changes requested):

Issue #32, opened 2018-05-17
CMS block coverage


paliarush:
As a Magento Developer, I want to query for CMS block content using the GraphQL API so that I can build basic ecommerce storefronts using only GraphQL for efficiency. 

**Use cases:** 
- Developer can query GraphQL for CMS block content and get back all necessary data to render the page on storefront or in 3rd party applications. 

**Acceptance Criteria**
- Only enabled blocks will be returned by the API
- Store view context is passed in the HTTP header and only blocks relevant to given context will be returned
- Staging versions of the page cannot be queried - only the currently active blocks are returned
- [DRAFT] API returns the following info
   - title
   - identifier
   - rendered (with no Magento-specific directives) or un-rendered block content, depending on configuration

Changes should be covered with web API functional tests. See [\Magento\GraphQl\Catalog\ProductViewTest::testQueryAllFieldsSimpleProduct](https://github.com/magento/magento2/blob/b3372064f574fd5aece0617009ee317db589febe/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/ProductViewTest.php#L35) as an example.

PR #109, opened 2018-06-30
#32 CMS block coverage

roma-glushko:
### Description
This PR enables CMS block API for GraphQL. Since it's merged, it would be possible to retrieve CMS blocks by identifier from GraphQL and the following information will be available:
- CMS block identifier
- CMS block title
- CMS block content

1. magento/graphql-ce#32: CMS block coverage


nuzil:
What about Status of Block? Can we extend to receive it also?

roma-glushko:
Hi @nuzil from the original requirements, the only enabled blocks can be retrieved by API. That's why status is not there. 

Can you please suggest a case when we need this? So we will be able to discuss that all together

nuzil:
@roma-glushko in this case agree

roma-glushko:
@nuzil however, I got the question. For the first iteration, there is no way to adjust list of available data to retrieve, but it will in further steps I guess. This is really something that we need to have here and everywhere in GraphQl API

PR #102, opened 2018-06-30
Fix the category tree depth calculation bug

dmytro-ch:
### Description
Fixed the issue with incorrectly calculated maximum level.

### Fixed Issues
1. magento/graphql-ce#100: Category tree depth calculation issue

### Manual testing scenarios
The result response should contain the last child item (specified on the deepest level according to the request).

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds on Travis CI are green)

Issue #100, opened 2018-06-30
Category tree depth calculation issue


dmytro-ch:
The maximum level value is being calculated incorrectly.

### Preconditions
The result response does not contain the last child item (specified on the deepest level according to the request).
Looks like the issue appears due to the incorrect maximum depth level calculation in ``\Magento\CatalogGraphQl\Model\Resolver\Products\DataProvider`` (line 100).

### Expected result
<img width="955" alt="screen shot 2018-06-30 at 12 26 38 pm" src="https://user-images.githubusercontent.com/11693779/42123848-8cb003a2-7c61-11e8-9ae9-d131802229d0.png">

### Actual result
<img width="949" alt="screen shot 2018-06-30 at 12 27 25 pm" src="https://user-images.githubusercontent.com/11693779/42123850-9197a230-7c61-11e8-922d-6c9a651c907e.png">

erikkubica:
@dmytro-ch as i see for your the category children is working, for some reason for me it´s not i have posted a issue here https://github.com/magento/graphql-ce/issues/123 can you see please anything work with my query?

naydav:
@erikkubica 
Hi, I will check https://github.com/magento/graphql-ce/issues/123 issue

naydav:
Merged with
https://github.com/magento/graphql-ce/pull/102

naydav:
One more PR
https://github.com/magento/graphql-ce/pull/121

PR #121, opened 2018-07-12
Fix regexp when filter by root category id

mhhansen:
### Description

At present, if you query the category tree requesting the Root Category (`id = 1`), query returns empty.
This is caused by the regular expression that filters the `catalog_category_entity`.`path` column table, which includes for all cases, a slash (`/`) at the beginning.
In the case of the Root Category, there's no slash at the beginning.

This has been tested for this scenarios:
* A fresh 2.3-alpha install, with no content created prior or after install.
* A fresh 2.3-alpha install, with other `Root Category` created post install.
* A fresh 2.3-alpha install, with sample-data created after installation.

For all cases, in the `catalog_category_entity`.`path` column table, categories are always created as a `1/` children.

*Fixed issue*: No issue created to this PR.
*Issue related*: Category tree depth calculation issue #100 

### Manual testing scenarios

To be able to test this PR, you have to first apply the PR #102, which fixes the category deph `children` param.

1. Query the `category(id: 1)`, to get the entire Category
2. Expected result: get the `Root Category` with all the children.
3. Actual result: 

### Query the Root Category
```
{
category(id: 1) {
    name
    path
    children_count
    children {
      id
      name
      level
      children {
        id
        name
      }
    }
  }
}
```

### Expected result
01-expected-result

### Actual result
02-actual-result

### Contribution checklist
 - [x] All new or changed code is covered with unit/integration tests (if applicable)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/graphql-ce?pullRequest=121) <br/>All committers have signed the CLA.

app/code/Magento/CatalogGraphQl/Model/Resolver/Products/DataProvider/CategoryTree.php
```diff
@@ -94,8 +94,17 @@ public function getTree(ResolveInfo $resolveInfo, int $rootCategoryId) : array
         $this->joinAttributesRecursively($collection, $categoryQuery);
         $depth = $this->depthCalculator->calculate($categoryQuery);
         $level = $this->levelCalculator->calculate($rootCategoryId);
+
+        // If root category is being filter, we've to remove first slash
+        if ($rootCategoryId == 1) {
```
naydav: Looks like we need to use `\Magento\Catalog\Model\Category::TREE_ROOT_ID` instead of `1`

mhhansen: that's exactly what I didn't found when I did that fix..
awesome! I will be pushing the update

mhhansen:
@naydav welcome back!
I remember we talk that this one would require to break the class down (or something like that),
to avoid coupling 13 deps in the `CategoryTree`. ping me at anytime!

magento-engcom-team:
@mhhansen thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

Issue #118, opened 2018-07-03
[Cart Operations] Create Empty Cart


paliarush:
Implement mutation for empty cart creation by the guest/customer user.

Cart ID is expected in the response. For the guest it should be hashed. 

**AC**:
- `createEmptyCart` mutation is declared. Should accept no arguments and return cart ID of type `String`. See example of [mutation declaration](https://github.com/magento/graphql-ce/blob/db3f2c73416c19f9cc13b1cbc48d504e843dc476/dev/tests/api-functional/_files/Magento/TestModuleGraphQlQuery/etc/schema.graphqls#L9)
- For guest (no token provided via `Authorization` header), the following service should be used `\Magento\Quote\Api\GuestCartManagementInterface::createEmptyCart`
- For customer (valid customer token is provided via `Authorization` header), the followigng service should be used `\Magento\Quote\Api\CartManagementInterface::createEmptyCartForCustomer`
- Changes are covered with web API functional tests. See [\Magento\GraphQl\TestModule\GraphQlMutationTest::testMutation](https://github.com/magento/graphql-ce/blob/db3f2c73416c19f9cc13b1cbc48d504e843dc476/dev/tests/api-functional/testsuite/Magento/GraphQl/TestModule/GraphQlMutationTest.php#L15) as an example.

naydav:
https://github.com/magento/graphql-ce/pull/166

PR #124, opened 2018-07-17
Empty cart creation added

rogyar:
### Description
Implementation of mutation for empty cart creation by the guest/customer user.
`createEmptyCart` mutation is declared which accepts no arguments. It returns a hashed cart ID for guest customers and the ~~real~~ hashed quote ID for logged in customers. The customer status (logged in/guest) is performed by checking authorization headers of the request.

1. https://github.com/magento/graphql-ce/issues/118 : [Mutations] Cart Operations > Create Empty Cart

### Manual testing scenarios
Use the following query to create an empty shopping cart:

```
mutation {
  createEmptyCart
}
```

To test with logged in customer, obtain the customer token and provide it in the request headers.


naydav (dismissed):
Looks like we need to add new module to
`composer.json` (and regenerate `composer.lock`)

paliarush:
@rogyar I have modified requirements slightly. We need to return cart ID hash for registered customers as well. Please let me know if you could do that change in current task, otherwise we will create a separate one.

Sorry for the change, but this decision was made after the task was created.

paliarush (approved):

app/code/Magento/QuoteGraphQl/Model/Resolver/Cart/CreateEmptyCart.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->valueFactory = $valueFactory;
+        $this->quoteIdMaskFactory = $quoteIdMaskFactory;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function resolve(Field $field, $context, ResolveInfo $info, array $value = null, array $args = null) : Value
+    {
+        $customerId = $context->getUserId();
```
paliarush: We should avoid working with global `$context` object, it is possible to inject `UserContextInterface` via constructor instead.

rogyar: That was my previous approach. But I saw that we already have `UserContextInterface` within a scope of $context and it doesn't make sense to inject it once again. 
Thank you for the tip, I will roll back the changes

paliarush (approved):

paliarush (approved):

PR #127, opened 2018-07-20
Fixed return value for getGraphQlClient in API-functional tests

rogyar:
### Description
`\Magento\TestFramework\TestCase\GraphQlAbstract::getGraphQlClient` is supposed to return `\Magento\TestFramework\TestCase\GraphQl\Client`. However, if `graphQlClient` was initialized previously, the method returns nothing. 
The fix adjusts the return behavior. So `getGraphQlClient` always returns correct value.

N/A

### Manual testing scenarios
N/A


dev/tests/api-functional/framework/Magento/TestFramework/TestCase/GraphQlAbstract.php
```diff
@@ -93,8 +93,8 @@ private function getGraphQlClient()
     {
         if ($this->graphQlClient === null) {
             return Bootstrap::getObjectManager()->get(\Magento\TestFramework\TestCase\GraphQl\Client::class);
-        } else {
-            $this->graphQlClient;
         }
+
```
naydav: Looks like logic should be

```
if ($this->graphQlClient === null) {
             $this->graphQlClient = Bootstrap::getObjectManager()->get(\Magento\TestFramework\TestCase\GraphQl\Client::class);
}
return $this->graphQlClient;
```

rogyar:
Hi @naydav. Thank you for your review. I've added the mentioned changes

Issue #88, opened 2018-06-06
'small_image' in 'ProductInterface' does not have enough info


DrewML:
### Preconditions

1. `2.3-develop` branch

### Steps to reproduce

1. Using `/graphql`, POST a query that requests the `small_image` for a product
2. The `string` returned for the `small_image` field will be in the format `/m/b/mb02-gray-0.jpg`.
3. The format returned does not have enough information for a client-rendered Magento application to append the uri to an `img` tag. The returned format is `https://mmnl.test/media/catalog/product/cache/2765542505660baab28ecd555e27366e/m/b/mb02-gray-0.jpg`

### Expected result

1. `small_image` (and other image values in the `ProductInterface`) should return a URL that can be used as-is to fetch the associated image

### Actual result

1. `small_image` is populated with a value that can not be used without obtaining other info about cache configuration.

### AC:
- Rename `small_image` to `small_image_url` and return URL in the following format `https://mmnl.test/media/catalog/product/cache/2765542505660baab28ecd555e27366e/m/b/mb02-gray-0.jpg`
- [Optional] Fix existing GraphQL tests. See [\Magento\GraphQl\Catalog\ProductViewTest::testQueryAllFieldsSimpleProduct](https://github.com/magento/magento2/blob/b3372064f574fd5aece0617009ee317db589febe/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/ProductViewTest.php#L35) as an example.

jissereitsma:
Perhaps there simply needs to come a second property `small_image_url` because you could argue that the image stays the same, while the URL might change depending on media URL settings. The problem here is that the GraphQL schema refers to the generic `ProductInterface` of the catalog module, which counts as `@api`. I've tried to fix things there, but that leads to a lot of work.

An easier solution would simply be to use a variable `media_url` within PWA? What I've now done for myself is create a new file `src/config.json` with a JSON string `{"media_url": "/media/catalog/product/"}`. Next, in the component where a real URL is needed, I import this JSON (`import config from 'src/config.json';`) and I construct the URL itself using `config.media_url + small_image`. Works for me.

DrewML:
> An easier solution would simply be to use a variable media_url within PWA? What I've now done for myself is create a new file src/config.json with a JSON string {"media_url": "/media/catalog/product/"}. Next, in the component where a real URL is needed, I import this JSON (import config from 'src/config.json';) and I construct the URL itself using config.media_url + small_image. Works for me.

[...]

PR #132, opened 2018-07-26
Small image URL added to the product data

rogyar:
### Description
This PR adds an extra field to the `ProductInterface` with image URL. We already have the `small_image` field with information about the image path. However, the image path information is not very useful for PWA applications that build a list of products and require rather URLs than some data that needs to be parsed.

1. https://github.com/magento/graphql-ce/issues/88 : 'small_image' in 'ProductInterface' does not have enough info

### Manual testing scenarios
1. Create a new product with/without an image
2. Make the following request:

```
{
  products (search: "withimages") {
    items {
      small_image_url
    }
  }
}
```
3. You should see the small image URL in the response:

```
{
  "data": {
    "products": {
      "items": [
        {
          "small_image_url": "http://docker.graphql.loc:8087/media/catalog/product/cache/0c40cc075c35680261029fa6d255feee/1/6/16262-11.jpg"
        }
      ]
    }
  }
}
```
4. If the requested product does not have an image, a placeholder URL should be returned


app/code/Magento/CatalogGraphQl/etc/schema.graphqls
```diff
@@ -259,6 +259,7 @@ interface ProductInterface @typeResolver(class: "Magento\\CatalogGraphQl\\Model\
     meta_description: String @doc(description: "A brief overview of the product for search results listings, maximum 255 characters")
     image: String @doc(description: "The relative path to the main image on the product page")
     small_image: String @doc(description: "The relative path to the small image, which is used on catalog pages")
+    small_image_url: String @doc(description: "The small image URL, which is used on catalog pages") @resolver(class: "Magento\\CatalogGraphQl\\Model\\Resolver\\Product\\SmallImageUrl")
```
vkublytskyi: Presence of both `small_image` and `small_image_url` may confuse developer. We should find another names or provide more details in the description. Also seems like similar pairs should be added for `image` and `thumbnail`. As an alternative solution, a path to media files may be part of a config type

paliarush: @rogyar how about having the following structure 
```
small_image: ProductImage
type ProductImage
{
   url: String
   path: String
}
```
 and then re-use it for other image types? 

@misha-kotov please provide a reference to the related tickets for other image types.

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
```diff
@@ -0,0 +1,131 @@
...
+            $result = function () {
+                return null;
+            };
+            return $this->valueFactory->create($result);
+        }
+        /** @var Product $product */
+        $product = $value['model'];
+
+        /* If small_image is not loaded for product, need to load it separately */
+        if (!$product->getSmallImage()) {
```
paliarush: What is the use case for it to be not loaded?

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
```diff
@@ -0,0 +1,131 @@
...
+            $productSmallImage = $this->getSmallImageFromGallery($productImages);
+            $product->setSmallImage($productSmallImage);
+        }
+
+        /* Design area is necessary to return the correct storefront image URL (or a placeholder) */
+        $area = $this->areaList->getArea(Area::AREA_FRONTEND);
+        $area->load(Area::PART_DESIGN);
+
+        $smallImageURL = $this->catalogImageHelper->init($product, 'product_small_image')->getUrl();
+        $product->getMediaAttributes();
```
paliarush: Why is this needed?

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
```diff
@@ -0,0 +1,131 @@
...
+            return $smallImageURL;
+        };
+
+        return $this->valueFactory->create($result);
+    }
+
+    /**
+     * Retrieves small image from the product gallery image
+     *
+     * @param $productImages
```
paliarush: Please specify argument type.

dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/MediaGalleryTest.php
```diff
@@ -0,0 +1,68 @@
...
+    items {
+   		small_image_url
+    }
+  }    
+}
+QUERY;
+        $response = $this->graphQlQuery($query);
+
+        self::assertArrayHasKey('small_image_url', $response['products']['items'][0]);
+        self::assertContains('placeholder/small_image.jpg', $response['products']['items'][0]['small_image_url']);
```
paliarush: Ideally we should verify that the image can be loaded using returned URL. Can be done just with curl request.

paliarush (changes requested):
Please address several comments above

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
rogyar: We need this part to generate a correct URL to the image. The product media gallery does not contain any information about the product image URL (only relative path). Upon a product image URL generation, the system takes into account some factors (like static content signing). The loaded area is also taken into account (in the core image retrieving flow). That's why we need to use these operations.
By removing the area loading, for example, we will always have a path (incorrect) to the placeholder instead of image URL.

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
rogyar: It's loaded when we specify both `small_image` and `small_image_url` in the GraphQl request. If we specify only `small_image_url` we don't have any information about the small image and cannot use `catalogImageHelper` for the image URL generation.

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
paliarush: Why we need `$product->getMediaAttributes()` call specifically? It looks like temporal coupling.

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
```diff
@@ -0,0 +1,131 @@
...
+        if (!$product->getSmallImage()) {
+            $galleryResource = $this->galleryResourceFactory->create();
+            $currentStoreId =$this->storeManager->getStore()->getId();
+            $productImages = $galleryResource->getProductImages($product, [$currentStoreId]);
+            $productSmallImage = $this->getSmallImageFromGallery($productImages);
+            $product->setSmallImage($productSmallImage);
+        }
+
+        /* Design area is necessary to return the correct storefront image URL (or a placeholder) */
+        $area = $this->areaList->getArea(Area::AREA_FRONTEND);
```
paliarush: We need to avoid area emulation. Instead we can copy configuration from storefront `di.xml` to graphql `di.xml`. @rogyar could you please check how much configuration need to be copied to make it work without area emulation?

app/code/Magento/CatalogGraphQl/etc/schema.graphqls
paliarush: Ticket for other image types https://github.com/magento/graphql-ce/issues/174

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
rogyar: Yes, you are right. It was an initial solution and I tried to reuse as much Magento core approaches as possible. Unfortunately, we have the temporal coupling here there in the core. I'm going to dig deeper and find more clear way for that. But, probably, we will need to introduce some new logic within a scope of this task for images URL generation that is different that we have in the code to avoid temporal coupling and other issues (especially performance issues).

app/code/Magento/CatalogGraphQl/Model/Resolver/Product/SmallImageUrl.php
rogyar: Sure. I will try to figure it out

paliarush (approved):

paliarush (approved):

naydav:
Failed tests
```
CategoryTest testCategoryProducts History | REST EE | < 1 sec
-- | -- | --
Magento\GraphQl\Catalog\CategoryTest::testCategoryProducts Exception: GraphQL response contains errors: Field "small_image" of type "ProductImage" must have a sub selection.   /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/framework/Magento/TestFramework/TestCase/GraphQl/Client.php:105 /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/framework/Magento/TestFramework/TestCase/GraphQl/Client.php:74 /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/framework/Magento/TestFramework/TestCase/GraphQlAbstract.php:49 /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/CategoryTest.php:262
````

```
ProductViewTest testQueryAllFieldsSimpleProduct History | REST EE | < 1 sec
-- | -- | --
Magento\GraphQl\Catalog\ProductViewTest::testQueryAllFieldsSimpleProduct Exception: GraphQL response contains errors: Field "small_image" of type "ProductImage" must have a sub selection.   /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/framework/Magento/TestFramework/TestCase/GraphQl/Client.php:105 /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/framework/Magento/TestFramework/TestCase/GraphQl/Client.php:74 /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/framework/Magento/TestFramework/TestCase/GraphQlAbstract.php:49 /opt/bamboo/builds/MCCE23-AT1210-RE/build-2/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/ProductViewTest.php:247
```

**Comment by rogyar** [comment]
I've just fixed the failing tests

**Comment by naydav** [comment]
We still have a problem with functionality
Some details
```
2018/09/17 18:11:53 ERROR - jmeter.util.BeanShellInterpreter: Error invoking bsh method: eval	Sourced file: inline evaluation of: ``String totalCount=vars.get("graphql_search_products_query_total_count");  if (to . . . '' : Method Invocation Integer.parseInt 
2018/09/17 18:11:53 WARN  - jmeter.assertions.BeanShellAssertion: org.apache.jorphan.util.JMeterException: Error invoking bsh method: eval	Sourced file: inline evaluation of: ``String totalCount=vars.get("graphql_search_products_query_total_count");  if (to . 
```

**Comment by naydav** [comment]
Need to fix `setup/performance-toolkit/benchmark.jmx`

**Review by paliarush — APPROVED** [review]

File: app/code/Magento/CatalogGraphQl/Model/Resolver/Product/Image.php [hunk 7]
```diff
@@ -43,7 +44,7 @@ public function resolve(
         array $args = null
     ): array {
         if (!isset($value['model'])) {
-            throw new \LogicException(__("Cannot resolve entity model"));
+            throw new GraphQlInputException(__('"model" value should be specified'));
```
paliarush: Why? Client knows nothing about the model. It must be passed from the parent resolver. I think before the change we had correct exception, the only change I would made is remove translation:
`throw new \LogicException("Cannot resolve entity model");`

naydav: https://github.com/magento/graphql-ce/issues/189
Has been created

paliarush (approved):

PR #140, opened 2018-08-02
Added HTML renderer for description and short description

rogyar:
### Description
This PR improves output for description and shord_description of a product. Previously we just had a raw output. Now, if you use some directive (upon inserting a block into the description or so), its contents are rendered.

### Fixed Issues (if relevant)
1. https://github.com/magento/graphql-ce/issues/134: GraphQL API returns rendered content for product

### Manual testing scenarios
1. Magento installation from `2.3-develop` branch (optionally: with sample data installed)
2. Add a static block to a product description or short description. Example (for sample data)

```
A plain text description is here ...
...
{{block id="sale-block"}}
```
3. Get the product's description and the short description using a GraphQL request.
4. You should see rendered values in the place where the block is inserted

PR #148, opened 2018-08-05
Render category description directives

rogyar:
### Description
This PR improves output for description of a category. Previously we just had a raw output. Now, if you use some directive (upon uploading an image), its contents are rendered

1. https://github.com/magento/graphql-ce/issues/135: GraphQL API returns rendered content for category

### Manual testing scenarios
1. Magento installation from 2.3-develop branch (optionally: with sample data installed)
2. Open a category and upload some image in the category description
3. Make a GraphQL request with category description included:

```
{
  category(id: 15) {
    name
    description
  }
}
```
4. In the description, you should see full image URL instead of `{media url=...}`

Please note, not all directives are supported in the category description on the core level. For example, you cannot insert a block like in https://github.com/magento/graphql-ce/pull/140 . It's not the current implementation limitation but core behavior. If you try to insert a block and check the storefront, it's not rendered there as well. All directives that are supported by default template filter (used on storefront) are also supported in the GraphQL response.


paliarush (approved):

paliarush:
@rogyar Thanks for contribution, changes look good!

melnikovi:
@rogyar I can not reproduce issue with not all directives being rendered on 2.3-develop, could you please provide steps you use to reproduce it? Thank you.

rogyar:
Hi @melnikovi, sure. Within a scope of a product description or short_description we can use different directives including inserting blocks/widgets etc. We cannot do the same with category description due to the core template renderer limitations. For example, if you insert in the product's description the following directive:
```
{{block id="sale-block"}}
```
it's rendered on the storefront (also, in the GraphQL response).
At the same time, if you try to insert the same directive into the category's description, it's not rendered on the storefront and the GraphQL response. So, I just put a note that we have some limitations from the template renderer side, which is not a sort of bug but the standard system behavior.

paliarush (approved):

PR #207, opened 2018-10-08
[GraphQL] Performance issue in filters realization

naydav:


lib/internal/Magento/Framework/GraphQl/Query/Fields.php
```diff
@@ -0,0 +1,64 @@
...
+                $queryAst,
+                [
+                    'leave' => [
+                        NodeKind::NAME => function (Node $node) use (&$queryFields) {
+                            $queryFields[$node->value] = $node->value;
+                        }
+                    ]
+                ]
+            );
+        } catch (\Exception $e) {
```
naydav, line 44: Why do we need empty catch?

naydav:
Merged with
https://github.com/magento/magento2ce/pull/3306

PR #209, opened 2018-10-12
Defined Constant for Url key

RakeshJesadiya:
Defined Constant for Url key in Page.php file.

### DIFF
--- a/app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
+++ b/app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
@@ -26,6 +26,11 @@ class Page
      * @var PageRepositoryInterface
      */
     private $pageRepository;
+    
+    /**
+     * Page Url Key
+     */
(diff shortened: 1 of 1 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/graphql-ce?pullRequest=209) <br/>All committers have signed the CLA.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -26,6 +26,11 @@ class Page
      * @var PageRepositoryInterface
      */
     private $pageRepository;
+    
+    /**
+     * Page Url Key
+     */
+    const URL_KEY = 'url_key';
```
naydav, line 33: Unfortunately, in this case, it makes no sense to use a constant. This value is used in just one place.
If you would like to contribute, please ping me(@vnayda ) in Slack channel and I will assist you with a suitable task

Issue #238, opened 2018-11-06
Configurable Products counted but not retrieved in GraphQL (works with REST endpoint)


koen-dev:
### Preconditions (*)

1. Magento Commerce 2.3.0 beta32
2. Custom instance with custom attributes and attribute sets

**Attributes used:**
* Growth method: Dropdown
* Height: Dropdown

### Steps to reproduce (*)

1. Create Product (Required fields + Categories)
2. Create Configurations for Product
3. Select Custom Attributes for Configurations
4. Select Custom Attributes Values
5. Skipped image uploading
6. Enter price for each configuration
7. Enter quantity for each configuration
8. Generate Products
9. Retrieve Category with products using GraphQL
```
{
  category(id: 5) {
    product_count
    products {
      items {
        name
      }
    }
  }
}
```

### Expected result (*)

1. Category with products as name and product_count
```
{
  "data": {
    "category": {
      "product_count": 1,
      "products": {
        "items": [
          {
            "name": "Buxushaag kant-en-klaar"
          }
        ]
      }
    }
  }
}
```

### Actual result (*)

1. Category with product_count matching Magento admin but without products
```
{
  "data": {
    "category": {
      "product_count": 1,
      "products": {
        "items": []
      }
    }
  }
}
```

### Extra information
This isn't only related to **CategoryTree** it also happens with **Products**.
When retrieving the product using the `REST API` it works as expected.

rogyar:
Please, doublecheck that at least one simple product is in stock and has qty greater than zero. 
This issue appears when the product is out of stock. The products count has a positive value but no products are on the list

koen-dev:
@rogyar I can confirm this has been resolved by adding quantity to everything.

rogyar:
Also, this behavior is not the final one. 
The product count value has been adjusted to reflect the fetched products number in the following PR

https://github.com/magento/graphql-ce/pull/243

Issue #214, opened 2018-10-15
Incorrect returned values of product_count


Serunde:
### Preconditions

1. Using Magento 2.3
2. Luma sample data installed

### Steps to reproduce

1. Form a category query including the `product_count` field with a category id of 3 (gear)
2. Include an inner product query with the `total_count` field.

### Expected result

`product_count` matches the total number of items returned from the query. Should be the same number as `total_count` if the inner product query does not have filters.

### Actual result

![productcount](https://user-images.githubusercontent.com/17077852/46960544-a55d0580-d053-11e8-854a-c0850f5a5c2f.png)

### Additional Notes
Worth mentioning that the value of `product_count` here is the same as the highest id in the category, despite the gap of products between id 23 and 36.

ronak2ram:
@Serunde 
I debug this issue and I found.

There is **product_count** show product's count which has a Visibility is **Catalog, Search** or **Not Visible Individually**
There is **total_count** show product's count which has a Visibility is **Catalog, Search**

And, Result items has products which has a Visibility is **Catalog, Search**

Maybe this help to mentor.

Serunde:
@ronak2ram 
Thanks for looking into this! So then this is the intended behavior of `product_count`?

misha-kotov:
@ronak2ram @Serunde 
Both fields should reflect products that have visibility set to `Catalog, Search` OR `Catalog`. Same should be reflected in the result set. 
This is because querying products in this way emulates user browsing products (as opposed to using search) and therefore products that would be reflected on storefront in this case should be returned by GraphQL.

ronak2ram:
@Serunde Yes
@misha-kotov 
We want ***product_count*** and ***total_count*** both should be same which has a Visibility is ***Catalog, Search*** only right?

misha-kotov:
@ronak2ram no, see my answer above. Products that are set to visibility `Catalog` should also be included.

naydav:
@TomashKhamlai 
Please, verify issue

TomashKhamlai:
#### Query:
```graphql
query category(
  $id: Int!
  $currentPage: Int!
  $pageSize: Int!
) {
  category(
    id: $id
  ) {
    name
    product_count
    products(
      currentPage: $currentPage
      pageSize: $pageSize
    ) {
      items {
        id
      }
      total_count
    }
  }
}
```

#### Variables:
```json
{
  "id": 3,
  "currentPage": 6,
  "pageSize": 6
}
```

[...]

PR #243, opened 2018-11-08
Added a separate resolver for returning category products count

rogyar:
### Description (*)
The `product_count` field of the category query now returns a value from the corresponding field of `catalog_category_entity`. Using this approach brings additional issues since the value in the database sometimes does not reflect the real products count. Moreover, if we request products of this category, the GraphQL query returns only products that are saleable (visible and in stock). So `product_count` value will show the different result if some products of the category are out of stock or not visible.

1. https://github.com/magento/graphql-ce/issues/214 : Incorrect returned values of product_count
2. https://github.com/magento/graphql-ce/issues/238 : Configurable Products counted but not retrieved in GraphQL (works with REST endpoint)

1. Add some products to a category and make some of them out of stock or not visible individually
2. Use a similar query to retrieve the category and products:

```
{
  category(id: 1158) {
    product_count
    products {
      items {
        name
      }
      total_count
    }
  }
}
```

3. The value in `product_count` field should corresponds to the total number of fetched products
4. The value in `total_count` field should equal the value in `products_count`

### DIFF
--- a/app/code/Magento/CatalogGraphQl/Model/Resolver/Category/ProductsCount.php
+++ b/app/code/Magento/CatalogGraphQl/Model/Resolver/Category/ProductsCount.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\CatalogGraphQl\Model\Resolver\Category;
+
+use Magento\Catalog\Model\Category;
+use Magento\Catalog\Model\Product\Visibility;
+use Magento\CatalogGraphQl\Model\Resolver\Products\DataProvider\Product\CollectionProcessor\StockProcessor;
+use Magento\Framework\Api\SearchCriteriaInterface;
+use Magento\Framework\GraphQl\Exception\GraphQlInputException;
+use Magento\Framework\GraphQl\Schema\Type\ResolveInfo;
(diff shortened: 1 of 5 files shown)


app/code/Magento/CatalogGraphQl/Model/Resolver/Category/ProductsCount.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public function resolve(Field $field, $context, ResolveInfo $info, array $value = null, array $args = null)
+    {
+        if (!isset($value['model'])) {
+            throw new GraphQlInputException(__('"model" value should be specified'));
+        }
+        /** @var Category $category */
+        $category = $value['model'];
+        $productsCollection = $category->getProductCollection();
+        $productsCollection->setVisibility($this->catalogProductVisibility->getVisibleInSiteIds());
```
paliarush, line 65: Ideally we should have some test cases for tricky scenarios, like:
- show products out of stock
- configurables with out of stock children
- products available in one store, but not available in the other

paliarush (approved):

app/code/Magento/CatalogGraphQl/Model/Resolver/Category/ProductsCount.php
naydav: Corresponding ticket has been created 
https://github.com/magento/graphql-ce/issues/255

PR #254, opened 2018-11-19
Create customer account functionality

Stepa4man:
### Description
Added the ability to create customer account through the GraphQL

### Fixed Issues
1. magento/graphql-ce/#248 Create a customer account

Run in GraphiQL:
`
mutation {
    createCustomer (
        input: {
            firstname: "Test"
            lastname: "Dev"
            email: "test@gmail.com"
            is_subscribed:true
        }
    ) {
        customer{
            id,
            firstname,
            lastname,
            email,
            is_subscribed
        }
    }
}
`


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/graphql-ce?pullRequest=254) <br/>All committers have signed the CLA.

Stepa4man:
@naydav, I haven't written tests yet. Will be covered a bit later.

naydav (changes requested):
- Need to add a possibility to set `customer EAV attributes` (include validation rules) 
Pay attention EAV attributes have own scope

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,130 @@
...
+        $this->dataObjectHelper->populateWithArray(
+            $customerDataObject,
+            $args['input'],
+            \Magento\Customer\Api\Data\CustomerInterface::class
+        );
+        $store = $this->storeManager->getStore();
+        $customerDataObject->setWebsiteId($store->getWebsiteId());
+        $customerDataObject->setStoreId($store->getId());
+
+        $password = array_key_exists('password', $args['input']) ? $args['input']['password'] : null;
```
naydav: When do we not need a password?

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,130 @@
...
+        array $value = null,
+        array $args = null
+    ) {
+        if (!isset($args['input']) || !is_array($args['input']) || empty($args['input'])) {
+            throw new GraphQlInputException(__('"input" value should be specified'));
+        }
+        try {
+            $customer = $this->createUserAccount($args);
+            $this->setUpUserContext($context, $customer);
+            if (array_key_exists('is_subscribed', $args['input'])) {
```
naydav, line 81: Pls, reuse `\Magento\CustomerGraphQl\Model\Customer\ChangeSubscriptionStatus`

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,130 @@
...
+     * @inheritdoc
+     */
+    public function resolve(
+        Field $field,
+        $context,
+        ResolveInfo $info,
+        array $value = null,
+        array $args = null
+    ) {
+        if (!isset($args['input']) || !is_array($args['input']) || empty($args['input'])) {
```
naydav, line 74: Need to have more smarty validation (based on `is_required` values of Customer EAV attributes)

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,130 @@
...
+            $customer = $this->createUserAccount($args);
+            $this->setUpUserContext($context, $customer);
+            if (array_key_exists('is_subscribed', $args['input'])) {
+                if ($args['input']['is_subscribed']) {
+                    $this->subscriberFactory->create()->subscribeCustomerById($customer->getId());
+                }
+            }
+            $data = $this->customerDataProvider->getCustomerById((int)$customer->getId());
+        } catch (LocalizedException $e) {
+            throw new GraphQlInputException(__($e->getMessage()));
```
naydav, line 90: We don't need to highlight to client all of possible errors

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,130 @@
...
+        }
+        return ['customer' => $data];
+    }
+
+    private function createUserAccount($args)
```
naydav: Pls, add dockblcoks and type hinting to all of methods (including return values)

naydav (changes requested):

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
Stepa4man: When we want to have registration and then send email to the customer with a link to set a password.
IN current state it works like that:
image

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
Stepa4man: Done

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
Stepa4man: I got it from the current \Magento\CustomerGraphQl\Model\Resolver\UpdateCustomer implementation. Is it ok for updating but not ok for creation? 

Just question to think. We can have attribute requirement changed between account creation and account updating. So it also needs to be validated in customer update.

My implementation checks which fields are required

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
Stepa4man: Done

Stepa4man:
@naydav eav attributes are implemented

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
Stepa4man: Fixed

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,160 @@
...
+use Magento\Framework\GraphQl\Exception\GraphQlInputException;
+use Magento\Framework\GraphQl\Query\Resolver\ContextInterface;
+use Magento\Framework\GraphQl\Query\Resolver\Value;
+use Magento\Framework\GraphQl\Query\ResolverInterface;
+use Magento\Framework\GraphQl\Schema\Type\ResolveInfo;
+use Magento\Framework\Validator\Exception as ValidatorException;
+use Magento\Newsletter\Model\SubscriberFactory;
+use Magento\Store\Model\StoreManagerInterface;
+
+class CreateCustomer implements ResolverInterface
```
naydav, line 24: Need to split this class (an example is in other resolvers of this module)

naydav, line 24: Need to create `dockblock` for all of classes

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,160 @@
...
+{
+    /**
+     * @var CustomerDataProvider
+     */
+    private $customerDataProvider;
+    /**
```
naydav, line 31: Missed rmpty line

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
```diff
@@ -0,0 +1,160 @@
...
+     * @param ContextInterface $context
+     * @param ResolveInfo $info
+     * @param array|null $value
+     * @param array|null $args
+     * @return array|Value|mixed
+     * @throws GraphQlInputException
+     * @throws LocalizedException
+     * @throws NoSuchEntityException
+     */
+    public function resolve(
```
naydav, line 67: We can use `inheritdoc`

app/code/Magento/CustomerGraphQl/Model/Resolver/CreateCustomer.php
naydav: `Magento\CustomerGraphQl\Model\Resolver\UpdateCustomer` does not contain any validation related to settings in admin panel (there is only one rule `need to provide password if you change email`)

naydav (changes requested):

misha-kotov:
@Stepa4man @naydav Thanks for the demo today! After going back to check on the use cases, it does not look like a guest shopper can create an account without a password on storefront. The Customer configuration options do now allow this either. The only way to create an account without a password is in Admin, which is a use case we do not support. So the password field should be required when creating an account through GraphQL.

Stepa4man:
@misha-kotov 
Misha, I still don't agree with you. You have to take into an account that it's just additional possibility for merchants to implement their own vision. 
It's not a security issue. 
Restricting this we will just restrict Magento's flexibility, I think.

Let me just give you some use case for example:
After a successful checkout guest customer will see a button to create an account. After hitting this button account will be created and the invitation to set a password will be sent to the customer email. So the customer can set this password at any time when it will be convenient for him... Not only right after order checkout.
Please consider my claim=)

misha-kotov:
After discussing this further, we decided to make the password field not required to allow for alternative account creation flows. So out of the box behavior of the endpoint would be: 
- if the password is specified, customer account is created right away with the password
- if the password is not specified, an email is sent to the customer with the `set password` link

magento-engcom-team:
@Stepa4man thank you for contributing. Please accept Community Contributors team invitation [here](https://github.com/orgs/magento/invitation) to gain extended permissions for this repository.

Issue #66, opened 2018-05-22
[Cart Operations] Fetch registered shopper cart


misha-kotov:
As a Magento developer, I need to fetch the shopping cart via GraphQL so that I can build basic ecommerce experiences for shoppers on the front-end using only GraphQL.

As an iteration of GraphQL support for cart, expose registered shopper cart query using resolver based on service contracts.

References:
- [How to expose a new query via GraphQL in Magento](https://devdocs.magento.com/guides/v2.3/graphql/develop/create-graphqls-file.html)
- TODO: specify equivalent registered shopper interface: 
[\Magento\Quote\Api\GuestCartRepositoryInterface::get](https://github.com/magento/magento2/blob/b2ce2a37d921b5ad88fc38663fc0ff3dd6c582d1/app/code/Magento/Quote/Api/GuestCartRepositoryInterface.php#L22)

**Acceptance Criteria**:
1. It is possible, using GraphQL, to query information about the cart created via REST APIs
1. Changes are covered with web API functional tests. See [\Magento\GraphQl\Catalog\ProductViewTest::testQueryAllFieldsSimpleProduct](https://github.com/magento/magento2/blob/b3372064f574fd5aece0617009ee317db589febe/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/ProductViewTest.php#L35) as an example.
1. Appropriate documentation is updated in DevDocs

naydav:
https://github.com/magento/graphql-ce/pull/266

Issue #8, opened 2018-05-09
[Cart Operations] Fetch guest shopper cart


paliarush:
As a Magento developer, I need to fetch the shopping cart via GraphQL so that I can build basic ecommerce experiences for shoppers on the front-end using only GraphQL.

As an initial iteration of GraphQL support for cart, expose guest cart query using resolver based on service contracts.

References:
- [How to expose a new query via GraphQL in Magento](https://devdocs.magento.com/guides/v2.3/graphql/develop/create-graphqls-file.html)
- [\Magento\Quote\Api\GuestCartRepositoryInterface::get](https://github.com/magento/magento2/blob/b2ce2a37d921b5ad88fc38663fc0ff3dd6c582d1/app/code/Magento/Quote/Api/GuestCartRepositoryInterface.php#L22)

**Acceptance Criteria**:
1. It is possible, using GraphQL, to query information about the cart created via REST APIs
1. Changes are covered with web API functional tests. See [\Magento\GraphQl\Catalog\ProductViewTest::testQueryAllFieldsSimpleProduct](https://github.com/magento/magento2/blob/b3372064f574fd5aece0617009ee317db589febe/dev/tests/api-functional/testsuite/Magento/GraphQl/Catalog/ProductViewTest.php#L35) as an example.
1. Appropriate documentation is updated in DevDocs

roma-glushko:
Hi @paliarush how are you?

I'm working on this task right now and I have a couple of questions I would like to clarify with you.
1. Do we really need to distinguish between guest/customer quotes? They are returned by different service contracts right now, but result interface is the same. Probably it may make easier to deal with both cases in the same way on the frontend. CC @nuzil 
2. I'm thinking about such signature for quote API:

```
type Query {
    quote(quoteId: String): Quote @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\QuoteResolver") @doc(description: "Quote")
}

type Quote @doc(description: "") {
    store_id: Int @doc(description: "")
    customer: Customer @doc(description: "")
    customer_note: String @doc(description: "")
    currency: CurrencyEnum @doc(description: "")
    is_guest: Boolean @doc(description: "")
    is_virtual: Boolean @doc(description: "")
    items_qty: Int @doc(description: "")
    items_count: Int @doc(description: "")
    billing_address: CustomerAddress @doc(description: "")
    items: [QuoteItem] @doc(description: "")
}

type QuoteItem @doc(description: "") {
[...]

PR #266, opened 2018-11-28
Query for retrieving shopping cart information

rogyar:
### Description (*)
This PR introduces the possibility to retrieve shopping cart information for guests and registered users by `cart_id`.

1. https://github.com/magento/graphql-ce/issues/8: [Query] Fetch guest shopper cart
2. https://github.com/magento/graphql-ce/issues/66: [Query] Fetch registered shopper cart

1. Create a shopping cart (optionally add items, set shipping method) and retrieve the shopping cart id (hashed)
2. Use the cart id to retrieve shopping cart information

```
{
  Cart(cart_id: "fda39cfe15aabde6d3c362cb3878de5s") {
    applied_coupon {
    	  code
    }
    items {
      id
    }
    addresses {
      firstname,
      lastname,
      address_type
    }
  }
}
```


dev/tests/integration/testsuite/Magento/Checkout/_files/active_quote.php
```diff
@@ -9,3 +9,11 @@
     ->setIsMultiShipping(false)
```
naydav, line 9: Please, add rollback for this fixture

dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/GetCartTest.php
```diff
@@ -0,0 +1,171 @@
+<?php /** @noinspection SpellCheckingInspection */
```
naydav: Looks like we need to remove this string

dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/GetCartTest.php
```diff
@@ -0,0 +1,171 @@
...
+        $this->graphQlQuery($query);
+    }
+
+    /**
+     * @magentoApiDataFixture Magento/Checkout/_files/active_quote.php
+     */
+    public function testGetCartForGuest()
+    {
+        $reservedOrderId = 'test_order_1';
+        $this->quoteResource->load(
```
naydav, line 103: Why do we need to load Quote?

rogyar: Let me shed some light here.
I use this fixture because we perform tests for this quote. I use this fixture since it assigns maskedQuoteId to the quote. Also, this fixture provides a guest quote (it's not assigned to a customer)
So, the flow is: we load the guest quote, we perform tests for the guest quote.

As for the model loading, we need to get maskedQuoteId somehow but we don't have a quoteId that can be used for quoteIdToMaskedId service (only reservedOrderId). So, I load the model for retrieving quoteId.

dev/tests/integration/testsuite/Magento/Checkout/_files/active_quote.php
rogyar: The rollback for the mentioned fixture exists (active_quote_rollback.php). As for masked id - the corresponding database entry is being removed when the quote is removed, so the current rollback works for masked ids as well (have just double checked this behavior)

dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/GetCartTest.php
rogyar: Thanks, it's PHPStorm stuff. Removed :)

Issue #232, opened 2018-11-02
GraphQL tools cannot perform "standard introspection query" in production mode


zetlen:
### Summary (*)
GraphQL tools for query validation and schema discovery rely on introspection features. While some tools perform many types of introspection query, or add introspection fields to arbitrary queries, the vast majority of tools perform a single, invariant [schema introspection](https://facebook.github.io/graphql/June2018/#sec-Schema-Introspection) query, described in the [reference implementation](https://graphql.org/graphql-js/utilities/#introspectionquery) as sufficient to download the entire type system.

All these tools utilize the [same introspection query](https://github.com/graphql/graphql-js/blob/master/src/utilities/introspectionQuery.js#L21-L111):
 - [GraphiQL](https://github.com/graphql/graphiql/blob/master/src/utility/introspectionQueries.js#L8)
 - [GraphQL Language Service](https://github.com/jimkyndemeyer/js-graphql-language-service/blob/master/src/project.js#L14), used by the PHPStorm / WebStorm GraphQL Plugin
 - [GraphQL Playground](https://github.com/prisma/graphql-playground/blob/master/packages/graphql-playground-react/src/components/Playground/SchemaFetcher.ts#L1), used in the PWA Studio Dev Server

However, Production instances of Magento 2.3 do not allow any introspection, including the standard query. This disables the tools for any user who does not have access to an instance of Magento 2 in developer mode. It's an obstacle to usability by new developers and community developers who are reliant on cloud instances.

### Examples (*)

Most public GraphQL APIs I could find permit the use of this standard introspection query:
- https://graphql.brandfolder.com/
- https://developer.github.com/v4/explorer/ (for logged-in users)
- https://developers.facebook.com/docs/graph-api/explorer/ (for logged-in developers)
- https://help.shopify.com/en/api/custom-storefronts/storefront-api/graphql-explorer

Production instances of Magento 2.3 return an error when running this query.

### Proposed solution
The standard introspection query is invariant, so it should be easy to cache its results. The GraphQL schema can only change when a new module is enabled or disabled, or when a new deploy occurs. That might require an on-demand refresh of the introspection query result, but it's still very safe to cache. Arbitrary introspection can still be disabled in production--but this common, standard query should be supported so that common, popular tools can download the schema and docs.

real34:
[...]

PR #308, opened 2018-12-31
Allow introspection by default in production mode

pmclain:
### Description (*)
Adds env.php parameter for disabling introspection:
```php
...
    'graphql' => [
        'disable_introspection' => true,
    ],
...
```

1. magento/graphql-ce#232: GraphQL tools cannot perform "standard introspection query" in production mode

#### Validate disabled
1. Enable production mode
2. Add the following to `env.php`:
```php
...
    'graphql' => [
        'disable_introspection' => true,
    ],
...
```
3. Query:
```graphql
{
  __schema {
    types {
      name
      interfaces {
        name
        description
      }
    }
  }
}
```
4. Expect error message `GraphQL introspection is not allowed, but the query contained __schema or __type`

#### Validate enabled
1. Enable production mode
2. Add the following to `env.php`, or ensure path `graphql/disable_introspection` is not set:
```php
...
    'graphql' => [
        'disable_introspection' => false,
    ],
...
```
3. Query:
```graphql
{
  __schema {
    types {
      name
      interfaces {
        name
        description
      }
    }
  }
}
```
4. Expect valid response

### DIFF
--- a/lib/internal/Magento/Framework/GraphQl/Query/QueryComplexityLimiter.php
+++ b/lib/internal/Magento/Framework/GraphQl/Query/QueryComplexityLimiter.php
@@ -11,6 +11,7 @@ use GraphQL\Validator\DocumentValidator;
 use GraphQL\Validator\Rules\DisableIntrospection;
 use GraphQL\Validator\Rules\QueryDepth;
 use GraphQL\Validator\Rules\QueryComplexity;
+use Magento\Framework\App\ObjectManager;
 
 /**
  * QueryComplexityLimiter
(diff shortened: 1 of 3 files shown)


paliarush (approved):

lib/internal/Magento/Framework/GraphQl/Query/QueryComplexityLimiter.php
```diff
@@ -53,7 +61,7 @@ public function __construct(
     public function execute(): void
     {
         DocumentValidator::addRule(new QueryComplexity($this->queryComplexity));
-        DocumentValidator::addRule(new DisableIntrospection());
+        DocumentValidator::addRule(new DisableIntrospection((int) $this->introspectionConfig->isIntrospectionDisabled()));
```
pmclain, line 64: @naydav I updated this line casting the bool to int after this change way pushed yesterday https://github.com/magento/graphql-ce/pull/308/commits/b6fe3ff39e221b6604b29753f5c378d93a063296 Passing anything other than `int 0` here disables introspection https://github.com/webonyx/graphql-php/blob/v0.12.6/src/Validator/Rules/DisableIntrospection.php#L31

PR #320, opened 2019-01-23
Add virtual products to cart

rogyar:
### Description (*)
This PR adds a possibility to add virtual products to the shopping cart

1. https://github.com/magento/graphql-ce/issues/145: [Cart Operations] Add virtual product to Cart

Use the following query with your own values to add a virtual product

```
mutation {
  addVirtualProductsToCart(
    input: {
      cart_id: "JJErKhk2dtVa7BOyJDxPH2fv0DcGZkA1", 
      cartItems: [
        {
          data: {
            sku: "virt", 
            qty: 2
          }          
        }
      ]
    }
  ) {
    cart {
      cart_id
      items {
        id
      }
    }
  }
}
```


app/code/Magento/QuoteGraphQl/etc/schema.graphqls
```diff
@@ -15,6 +15,7 @@ type Mutation {
     setBillingAddressOnCart(input: SetBillingAddressOnCartInput): SetBillingAddressOnCartOutput
     setShippingMethodsOnCart(input: SetShippingMethodsOnCartInput): SetShippingMethodsOnCartOutput @resolver(class: "\\Magento\\QuoteGraphQl\\Model\\Resolver\\SetShippingMethodsOnCart")
     addSimpleProductsToCart(input: AddSimpleProductsToCartInput): AddSimpleProductsToCartOutput @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\AddSimpleProductsToCart")
+    addVirtualProductsToCart(input: AddVirtualProductsToCartInput): AddVirtualProductsToCartOutput @resolver(class: "Magento\\QuoteGraphQl\\Model\\Resolver\\AddVirtualProductsToCart")
```
rogyar: Could we reuse resolver from `addSimpleProductsToCart` (rename in order to have a more universal name) here? Resolvers for simple and virtual products are identical.

app/code/Magento/QuoteGraphQl/Model/Resolver/AddVirtualProductsToCart.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->addProductsToCart = $addProductsToCart;
+        $this->extractDataFromCart = $extractDataFromCart;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function resolve(Field $field, $context, ResolveInfo $info, array $value = null, array $args = null)
+    {
+        $cartHash = $this->arrayManager->get('input/cart_id', $args);
```
rogyar: If we are not going to use the resolver for virtual/simple products, should we add an additional check for the product type in a resolver?
Currently we can add simple product via virtual product mutation and vice versa

rogyar (commented):
@naydav Could you check my questions in the review, please? Thanks

app/code/Magento/QuoteGraphQl/etc/schema.graphqls
naydav: Looks like we need to reuse configurable for simple because the code is the same (and we will have copy paste detector fail)

app/code/Magento/QuoteGraphQl/Model/Resolver/AddVirtualProductsToCart.php
naydav: We need to have 2 different resolvers with checking of product type, but all general logic should be moved to Model in a separate class

Also, need to cover functionality with API-functional tests (including negative scenarios with a wrong product type)
Thanks

keharper:
Documented as part of https://github.com/magento/devdocs/pull/4502/

Issue #352, opened 2019-02-06
Add query isEmailAvailable on customer registration


VitaliyBoyko:
The REST API endpoint ```/rest/V1/customers/isEmailAvailable``` should be covered by GraphQl.

naydav:
https://github.com/magento/graphql-ce/pull/356

PR #356, opened 2019-02-07
graphQl-352: is email available

VitaliyBoyko:
### Description (*)
Is email available coverage.

Query:
```
{
  isEmailAvailable(email: "customer@example.com") {
    is_email_available
  }
}
```

Response:
```
{
  "data": {
    "isEmailAvailable": {
      "is_email_available": true
    }
  }
}
```

1. magento/graphql-ce#352: Add query isEmailAvailable on customer registration


app/code/Magento/CustomerGraphQl/Model/Customer/IsEmailAvailableDataProvider.php
```diff
@@ -0,0 +1,40 @@
...
+declare(strict_types=1);
+
+namespace Magento\CustomerGraphQl\Model\Customer;
+
+use Magento\Customer\Api\AccountManagementInterface;
+
+/**
+ * Is Customer Email Available checker
+ */
+class IsEmailAvailableDataProvider
```
naydav: Hello @VitaliyBoyko,
Looks like in this case we don't need a class wrapper 
we can call `AccountManagementInterface` directly in our resolver
Thanks

naydav (changes requested):

keharper:
Documented in https://github.com/magento/devdocs/pull/4556

PR #363, opened 2019-02-12
Always return cart id null after add product

dominicfernando:
Always return cart id null after add product

### Description (*)
When process below json then always return cart_id as null
```
mutation {
  addSimpleProductsToCart(input: {
        cart_id: "Dqghs1ZMSN4g0ayaNC7Hudy4xa9hGdAE",
        cartItems: {
            data: {
                    sku: "VA11-GO-NA",
                    qty:1
            }
      }
}) {
    cart {
      cart_id,
      items {
        id,
      qty
    }
    }
  }
}
```
### Before Fix
selection_017

### After Fix

selection_018


app/code/Magento/QuoteGraphQl/Model/Resolver/AddSimpleProductsToCart.php
```diff
@@ -81,6 +81,7 @@ public function resolve(Field $field, $context, ResolveInfo $info, array $value
 
         $this->addProductsToCart->execute($cart, $cartItems);
         $cartData = $this->extractDataFromCart->execute($cart);
+        $cartData["cart_id"] = $cartHash;
```
naydav: Hello @dominicfernando,
It's should be part of `ExtractDataFromCart` 
If `ExtractDataFromCart` will return `cart_id` then we will cover all of the cases (like adding different types of products)

We already have this fix in PR
https://github.com/magento/graphql-ce/pull/320/files#diff-b5e7e8d0c42394996dc70fa755840759R61 (will be included in 2.3.2)

You can create API-functional tests for this case

PR #421, opened 2019-03-04
GraphQl-416: graphql-input provides to Customer as many errors as appeared instead of last one like on Magento Storefront

XxXgeoXxX:
### Description (*)
Issue: https://github.com/magento/graphql-ce/issues/416

#### Preconditions (*)
Any supported product with more then 1 required customizable options
GraphQL:
```
{
  "errors": [
    {
      "message": "The product's required option(s) weren't entered.
Make sure the options are entered and try again.",
      "category": "graphql-input",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "addSimpleProductsToCart"
      ]
    }
  ],
  "data": {
    "addSimpleProductsToCart": null
  }
}
```


XxXgeoXxX:
Graph-ql return originl error string instead add message logic in
app/code/Magento/Checkout/Controller/Cart/Add.php:execute:155

app/code/Magento/QuoteGraphQl/Model/Cart/AddSimpleProductToCart.php
```diff
@@ -87,7 +87,8 @@ public function execute(Quote $cart, array $cartItemData): void
         }
 
         if (is_string($result)) {
-            throw new GraphQlInputException(__($result));
+            $uniqueMessqges = array_unique(explode("\n", $result));
```
naydav: Generally, need to investigate why do we have not unique messages... looks like a bug of Cart

XxXgeoXxX: Changes were reverted. 
FIx  was applied inbase level.

app/code/Magento/Catalog/Model/Product/Type/AbstractType.php
```diff
@@ -603,7 +603,7 @@ protected function _prepareOptions(\Magento\Framework\DataObject $buyRequest, $p
                 }
             }
             if (count($results) > 0) {
-                throw new LocalizedException(__(implode("\n", $results)));
+                throw new LocalizedException(__(implode("\n", array_unique($results))));
```
naydav: https://github.com/magento/graphql-ce/blob/1f8c1f301f0e3083bc5d64cbdde021876995e381/app/code/Magento/Catalog/Model/Product/Type/AbstractType.php#L596

vpodorozh: Hey @naydav , 

I see what do you mean, however on Magento storefront current behaviour is done on _presentation_ layer in `app/code/Magento/Checkout/Controller/Cart/Add.php:155` - and yes it is some kind of a duck-tape :)

But still, I do suggest to do the same - create the same "duck-tape" on the same _presentation_ layer here `app/code/Magento/QuoteGraphQl/Model/Cart/AddSimpleProductToCart.php:84` - in GraphQl.
And create a separate issue in Magento core to solve the problem with multiple messages from `app/code/Magento/Catalog/Model/Product/Type/AbstractType.php` (_domain_ layer).

Moreover, we are not able to make fixes of Magento core in GraphQl project according to this [article](https://github.com/magento/graphql-ce/wiki/Approach-to-Implementation#technical-requirements-to-features), so the current solution will be the best I suppose.

@naydav - What do you think about it?
Thx!

vpodorozh: @paliarush - could you check my prev comment?
Thx!
Maybe you could advise.

naydav: > Hey @naydav ,
> 
> I see what do you mean, however on Magento storefront current behaviour is done on _presentation_ layer in `app/code/Magento/Checkout/Controller/Cart/Add.php:155` - and yes it is some kind of a duck-tape :)
> 
> But still, I do suggest to do the same - create the same "duck-tape" on the same _presentation_ layer here `app/code/Magento/QuoteGraphQl/Model/Cart/AddSimpleProductToCart.php:84` - in GraphQl.
> And create a separate issue in Magento core to solve the problem with multiple messages from `app/code/Magento/Catalog/Model/Product/Type/AbstractType.php` (_domain_ layer).
> 
> Moreover, we are not able to make fixes of Magento core in GraphQl project according to this [article](https://github.com/magento/graphql-ce/wiki/Approach-to-Implementation#technical-requirements-to-features), so the current solution will be the best I suppose.
> 
> @naydav - What do you think about it?
> Thx!

Agree. Because the current fix is BIC.
Thanks for the review.

vpodorozh:
@XxXgeoXxX - any update on this PR?

XxXgeoXxX:
@vpodorozh 
after the discussion - "on Hold"

naydav:
@XxXgeoXxX @vpodorozh 
We can't change the current implementation because it will be BIC.
But what do you think about:

1) Introduce new Exception class like `InvalidOptionInput extends LocalizedException`. A new exception will take also product SKU and option id.

Something like:
```
if (count($extendedData) > 0) {
                throw new InvalidOptionInput(__(implode("\n", $results)), $extendedData);
            }
```
In this case it will not be BIC.

2) Take additional data from exception in GraphQL module and show as list of more detailed messages,.
Thanks

vpodorozh:
> @XxXgeoXxX @vpodorozh
> We can't change the current implementation because it will be BIC.
> But what do you think about:
> 
> 1. Introduce new Exception class like `InvalidOptionInput extends LocalizedException`. A new exception will take also product SKU and option id.
> 
> Something like:
> 
> ```
> if (count($extendedData) > 0) {
>                 throw new InvalidOptionInput(__(implode("\n", $results)), $extendedData);
>             }
> ```
> 
> In this case it will not be BIC.
> 
> 1. Take additional data from exception in GraphQL module and show as list of more detailed messages,.
>    Thanks

@naydav - for me it looks like an overengineering. Why don't we make the fix in the Magento core (new PR in core) and cancel current PR/issue?
Do you see the other usages of the proposed solution, other than for invalid options exceptions?

@XxXgeoXxX - what is your opinion?

Thx!

naydav:
> > @XxXgeoXxX @vpodorozh
> > We can't change the current implementation because it will be BIC.
> > But what do you think about:
> > 
> > 1. Introduce new Exception class like `InvalidOptionInput extends LocalizedException`. A new exception will take also product SKU and option id.
> > 
> > Something like:
> > ```
> > if (count($extendedData) > 0) {
> >                 throw new InvalidOptionInput(__(implode("\n", $results)), $extendedData);
> >             }
> > ```
> > 
> > 
> > In this case it will not be BIC.
> > 
> > 1. Take additional data from exception in GraphQL module and show as list of more detailed messages,.
> >    Thanks
> 
> @naydav - for me it looks like an overengineering. Why don't we make the fix in the Magento core (new PR in core) and cancel current PR/issue?
> Do you see the other usages of the proposed solution, other than for invalid options exceptions?
> 
> @XxXgeoXxX - what is your opinion?
> 
> Thx!


@vpodorozh 
Of course, it will be overengineering :) but we need to keep backward compatibility.
Changing of exception message is backward incompatible changes. That's why we need to have some additional code.

vpodorozh:
@XxXgeoXxX - let's do it!
Check the latest comment from @naydav with explanation, contact me in case you need some support.
Thx!

vpodorozh:
@XxXgeoXxX has committed to pick-up this issue with the proposed solution.

vpodorozh:
@novikor - please pick up this PR.
I do not have permissions for approval any more. Thx!

XxXgeoXxX:
@naydav 
I updated the solution according to disscussion.
Could we skip the static test file for app/code/Magento/Catalog/Model/Product/Type/AbstractType.php?

naydav:
> @naydav
> I updated the solution according to disscussion.
> Could we skip the static test file for app/code/Magento/Catalog/Model/Product/Type/AbstractType.php?

@XxXgeoXxX 
Yes, we can ignore it
thanks

naydav:
@TomashKhamlai please verify the fix
Thanks

TomashKhamlai:
Now it looks better because there is an explanation for each problem.

naydav:
`@see `\Magento\Framework\Exception\AggregateExceptionInterface``

lenaorobei:
@magento run all tests

lenaorobei:
@magento run all tests

lenaorobei:
@magento run all tests

naydav:
Hello @XxXgeoXxX ,
PR was closed due to inactivity (ticket was unassigned). Please reopen and update if you wish to continue.
Thanks!

lenaorobei:
Please address @naydav comment about the use of `Magento\Framework\Exception\AggregateExceptionInterface`.

lenaorobei:
Closing this PR in favor of https://github.com/magento/graphql-ce/issues/970
Should be fixed globally.

Issue #432, opened 2019-03-04
Test coverage: 'reversed quote id to masked cart id'


naydav:
Introduce class with logic `reversed quote id to masked cart id`

Replace all of the boilerplate code like
```
/**
     * @param string $reversedQuoteId
     * @return string
     */
    private function getMaskedQuoteIdByReversedQuoteId(string $reversedQuoteId): string
    {
        $quote = $this->quoteFactory->create();
        $this->quoteResource->load($quote, $reversedQuoteId, 'reserved_order_id');

        return $this->quoteIdToMaskedId->execute((int)$quote->getId());
    }
```
on usage on the new class (can get via object manager in setUp)

naydav:
https://github.com/magento/graphql-ce/pull/449

PR #449, opened 2019-03-06
GraphQL-432: Test coverage: 'reversed quote id to masked cart id'

AleksLi:
### Description (*)
Created separate class GetMaskedQuoteIdByReversedQuoteId and replaced method in one api-functional test.

1. magento/graphql-ce#432: Test coverage: 'reversed quote id to masked cart id'

1. Please run the api-functional tests for `\Magento\GraphQl\Quote\SetShippingAddressOnCartTest` class


app/code/Magento/QuoteGraphQl/Model/GetMaskedQuoteIdByReversedQuoteId.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+declare(strict_types=1);
+
+namespace Magento\QuoteGraphQl\Model;
+
+use Magento\Quote\Model\ResourceModel\Quote as QuoteResource;
+use Magento\Quote\Model\QuoteIdToMaskedQuoteIdInterface;
+use Magento\Quote\Model\QuoteFactory;
+
+class GetMaskedQuoteIdByReversedQuoteId
```
naydav: Please, reuse this class in all of the places (find by `getMaskedQuoteIdByReversedQuoteId`)
But not just in one testcase class

Also, this class is used just for testing, so need to move to test directory `dev/tests/api-functional/testsuite/Magento/GraphQl/Quote`
``

naydav (changes requested):

app/code/Magento/QuoteGraphQl/Model/GetMaskedQuoteIdByReversedQuoteId.php
atwixfirster: @naydav , I am sorry but I have a question about **GetMaskedQuoteIdByReversedQuoteId** class name. 

We load quote by `reserved_order_id` field
```
$this->quoteResource->load($quote, $reversedQuoteId, 'reserved_order_id');
```
and in the same time new class is called as `GetMaskedQuoteIdByReversedQuoteId`. 

May we should use a proper name for it - `GetMaskedQuoteIdByReservedOrderId`?

Thank you!

CC: @AleksLi

naydav: Hi @atwixfirster,
I saw the fix in your PR https://github.com/magento/graphql-ce/pull/450/files

But for avoiding a lot of merge conflicts I am going to merge all of related PRs and finally make the fix in your PR which will be merged last.

AleksLi: I didn't cover everything in all the places, because I thought it might be something with the class name or so. I will do that now

Issue #431, opened 2019-03-04
Test coverage: Using the "use_for_shipping" option with multishipping is not possible.


naydav:
Need to add https://github.com/magento/graphql-ce/blob/2.3-develop/app/code/Magento/QuoteGraphQl/Model/Cart/SetBillingAddressOnCart.php#L95 case to:

```
\Magento\GraphQl\Quote\Customer\SetBillingAddressOnCartTest
\Magento\GraphQl\Quote\Guest\SetBillingAddressOnCartTest
```

Examples you could find in
https://github.com/magento/graphql-ce/tree/2.3-develop/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Customer

https://github.com/magento/graphql-ce/tree/2.3-develop/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Guest

naydav:
Was covered in the scope of
https://github.com/magento/graphql-ce/pull/321
https://github.com/magento/graphql-ce/pull/330

PR #330, opened 2019-01-29
Prototype: setPaymentMethodOnCart

pmclain:
### Description (*)
Implements mutation for setting cart payment method. Additionally provides a convention for handling the dynamic `additional_data` object currently sent with REST payment information.  

The `input PaymentMethodInput` provides payment integrators the ability to define additional inputs for populating `additional_data`. I've included a partial example using the Braintree integration https://github.com/magento/graphql-ce/compare/2.3-develop...pmclain:issue/293?expand=1#diff-9cca29ccf0ab73e1ac28e56811a208a8.

Custom input types can converted to `additional_data` by registering a `Magento\QuoteGraphQl\Model\Cart\Payment\AdditionalDataBuilderInterface` for the payment method with the `Magento\QuoteGraphQl\Model\Cart\Payment\AdditionalDataBuilderPool`. The default assumption is a one-to-one match of input property to the desired array set as the payment method additional data.

Payment methods like Check/MO that do not have a dependency on the `additional_data` property work as-is and do not require registering an  `AdditionalDataBuilderInterface`

### Fixed Issues (if relevant)
1. #293 
Add support for Payment Methods:
#286 - Check/Money Order
#287 - Bank Transfer
#288 - Cash on Delivery
#290 - Purchase Order
Braintree and Braintree Vault (requires test coverage)

### Manual testing scenarios (*)
1. Create customer token
2. Create empty cart for customer
3. Using the customer token and masked quote id set payment method on cart with the mutation below:
```graphql
mutation {
  setPaymentMethodOnCart(input: 
    {
      cart_id: "{{maskedQuoteId}}", 
      payment_method: {
          method: "{{methodCode}}"
        }
      }) {
    
    cart {
      cart_id,
      payment_method {
        method
      }
    }
  }
}
```

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #321, opened 2019-01-23
GraphQL 296 set billing address

VitaliyBoyko:
This PR covers setting billing address on the cart.

### Fixed Issues (if relevant)
1. magento/graphql-ce#296: [Cart Operations] Set Billing Address

Example query:

 ```
 mutation {
  setBillingAddressOnCart(
    input: {
      cart_id: "maskedQuoteId"
      billing_address: {
        address: {
          firstname: "test firstname"
          lastname: "test lastname"
          company: "test company"
          street: ["test street 1", "test street 2"]
          city: "test city"
          region: "test region"
          postcode: "887766"
          country_code: "US"
          telephone: "88776655"
          save_in_address_book: false
        }
        use_for_shipping: true
      }
    }
  )
}
```

PS: PR based on my previous work (implementing multi-shipping).
In order to see changes properly, I've done PR to the multi-shipping branch.
I gonna change the base branch since multi-shipping be merged.

PR #452, opened 2019-03-07
GraphQL-431: Test coverage: Using the "use_for_shipping" option with multishipping is not possible

AleksLi:
Covered SetBillingAddressOnCartTest with multiple shipping addresses in a quote. Added Fixture for that.

1. magento/graphql-ce#431: Test coverage: Using the "use_for_shipping" option with multishipping is not possible

1. Run the tests of SetBillingAddressOnCartTest for all the customer types Guest|Saved Customer.


dev/tests/integration/testsuite/Magento/Checkout/_files/quote_with_multiple_addresses_saved.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
naydav: We already the same fixture
Please, check the same class, it should be used in another testcases

BTW if we create fixture we MUST create rollback
Thanks

dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Customer/SetBillingAddressOnCartTest.php
```diff
@@ -403,6 +403,51 @@ public function testSetBillingAddressIfCustomerIsNotOwnerOfAddress()
         $this->graphQlQuery($query, [], '', $this->getHeaderMap('customer2@search.example.com'));
```
naydav: Need to add the same case for Guest
https://github.com/magento/graphql-ce/blob/3469f4b251cdc4874048dfff0374f818cb757964/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Guest/SetBillingAddressOnCartTest.php

naydav (changes requested):

dev/tests/integration/testsuite/Magento/Checkout/_files/quote_with_multiple_addresses_saved.php
AleksLi: @naydav I haven't found it. Can you take a look and give me the name. I did look at the whole project.

dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Customer/SetBillingAddressOnCartTest.php
AleksLi: @naydav Yes, I will do that

dev/tests/integration/testsuite/Magento/Checkout/_files/guest_quote_with_multiple_addresses_saved.php
```diff
@@ -0,0 +1,61 @@
+<?php
```
naydav: Please, reuse current fixtures files
You can find examples for all of the cases in

https://github.com/magento/graphql-ce/tree/2.3-develop/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Customer

https://github.com/magento/graphql-ce/tree/2.3-develop/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Guest
(also in the same testcase class)

naydav (changes requested):

dev/tests/integration/testsuite/Magento/Checkout/_files/guest_quote_with_multiple_addresses_saved.php
AleksLi: @naydav  ok. I'll do that. Thanks.

naydav:
@AleksLi 
Implementation is overhead. You should not use create or modify quote or address model

What need to do:
1) Use the `quote with addresses` fixture as a precondition (could be found in the same file)
2) During test try to assign billing address with `use_for_shipping=true` flag
3) Check error message

Thanks

AleksLi:
@naydav please take a look on this one again. I know that some of the checks are FAILED but I don't know why, actually. I've seen the reason and it looks weird as for me.

naydav:
Was covered in the scope of
https://github.com/magento/graphql-ce/pull/321
https://github.com/magento/graphql-ce/pull/330

PR #461, opened 2019-03-11
GraphQl-407: mutation createEmptyCart doesn't throw exception when authorization header is set and token has expired

XxXgeoXxX:
### Description (*)
Issue https://github.com/magento/graphql-ce/issues/407

#### Steps to reproduce (*)
Generate Customer token
Wait for token expiration
Create empty cart using expired token


TomashKhamlai:
Hi @XxXgeoXxX. Have you tried to test this? I mean, could you describe the steps as a way to reproduce the problem that you have fixed? I would like to create a manual test.

And please try these steps also.
1) Checked that cron is running and crontab is configured to run every 2 minutes:
```crontab
#~ MAGENTO START 2bb453e2ff349dc1691dc37aae6670cd
*/2 * * * * /usr/bin/php7.1 /var/www/html/m2-test/bin/magento cron:run 2>&1 | grep -v "Ran jobs by schedule" >> /var/www/html/m2-test/var/log/magento.cron.log
*/2 * * * * /usr/bin/php7.1 /var/www/html/m2-test/update/cron.php >> /var/www/html/m2-test/var/log/update.cron.log
*/2 * * * * /usr/bin/php7.1 /var/www/html/m2-test/bin/magento setup:cron:run >> /var/www/html/m2-test/var/log/setup.cron.log
#~ MAGENTO END 2bb453e2ff349dc1691dc37aae6670cd
```
2) Applied this patch to set Token TTL to 6 minutes
```diff
Index: app/code/Magento/Integration/Helper/Oauth/Data.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- app/code/Magento/Integration/Helper/Oauth/Data.php	(date 1552294247000)
+++ app/code/Magento/Integration/Helper/Oauth/Data.php	(date 1552387994000)
@@ -121,7 +121,7 @@
     public function getCustomerTokenLifetime()
     {
         $hours = (int)$this->_scopeConfig->getValue('oauth/access_token_lifetime/customer');
-        return $hours > 0 ? $hours : 0;
+        return 0.1;
     }
 
     /**
```
3) Generated token using `generateCustomerToken`:
```graphql
"token": "e906giw9cwlfwifcl5g05lhbnliw6vz4"
```
4) Checked that token is in the database:

| entity_id | consumer_id | admin_id | customer_id | type | token | secret | verifier | callback_url | revoked | authorized | user_type | created_at |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 17 | NULL | NULL | 1 | access | e906giw9cwlfwifcl5g05lhbnliw6vz4 | n2odn4havjqycduz6o6qdbd05uuaczrl | NULL |  |  0  |  0  |  3  |  2019-03-12 13:28:18  |

5) Created empty cart
"createEmptyCart": "UMwDEOciIMdmvWyFtBnItuyC2gLWeClT"
(Actually returned active cart that has products, but this is the case too)

6) If there is some time repeat step 5) and verify that returned data was not modified
7) Wait for token expiration
8) Repeat step 5)

**Expected result:**
Guest cart is not created because token was used

**Actual result:**
Guest cart is created

XxXgeoXxX:
Hi, @TomashKhamlai 
Thank you for your test cases. I'm working on this issue and  I'll update my solution

TomashKhamlai:
@XxXgeoXxX, is this solution ready for testing?

XxXgeoXxX:
HI
@TomashKhamlai
I'm working on solution according to 'Magento way' which has BCI

TomashKhamlai:
Great! When solution will be ready put label "Ready for QA"

app/code/Magento/Webapi/Model/Authorization/TokenUserContext.php
```diff
@@ -179,7 +179,9 @@ protected function processRequest()
         if (!$token->getId() || $token->getRevoked() || $this->isTokenExpired($token)) {
             $this->isRequestProcessed = true;
 
-            return;
+            throw new \Magento\Framework\Exception\NoSuchEntityException(
```
naydav: @XxXgeoXxX As we considered in our last talk, the changes are backward incompatible
Need to revert it (including tests) and move logic to GraphQL responsible area (check place where user id is set to graphql context)

naydav (changes requested):

XxXgeoXxX:
Changes were reverted

naydav:
> Changes were reverted

Thanks
Next step is "move logic to GraphQL responsible area (check place where user id is set to graphql context)"

TomashKhamlai:
Now it works like Customers can generate so many Carts as they wish. But it seems to me that Customer should have only one active cart.

### Steps to reproduce:
1. Create Customer from Storefront
2. Generate Customer token
3. Add some product to Cart from Storefront
4. Create Empty Cart using Customer's token (actually, return the active Cart with product in it)
5. Perform query using Customers token.
#### Query:
```graphql{
  cart(cart_id: "<cart_id>") {
    items {
      id
    }
  }
}
```
#### Headers:
```json
{"Aythorization": "Bearer <token>"}
```

### Expected result:
Cart is not empty

### Actual result:
```graphql
{
  "data": {
    "cart": {
      "items": []
    }
  }
}
```

Please comment if it works for you as expected. I've reinstalled my testing environment, so a little chance is that it doesn't work only for me.

lenaorobei:
Closing this PR because unfortunately it does not solve the issue from a global perspective.

Issue #464, opened 2019-03-11
Currency misspelled in graphql attributes


erikmarr:
### Preconditions (*)
In the following PR: https://github.com/magento/graphql-ce/pull/313/files 'currency' is spelled 'currecy'. This occurs in the schema and test

### Steps to reproduce (*)
View the currency attributes in graphiql.

### Expected result (*)
'currecy' should be 'currency'

naydav:
https://github.com/magento/graphql-ce/pull/465

PR #398, opened 2019-02-25
Changed canonical_url to relative_url

yogeshsuhagiya:
### Description (*)
As per the devdoc, there must be a **relative_url** instead of **canonical_url**.
- https://devdocs.magento.com/guides/v2.3/graphql/reference/url-resolver.html

### Fixed Issues (if relevant)
1. N/A

### Manual testing scenarios (*)
1. Try to execute following GraphQl query as explain in docs.
`{
 urlResolver(url:"joust-duffle-bag.html") {
   id
   relative_url
   type
 }
}`
2. It throws following error:
`{
  "errors": [
    {
      "message": "Cannot query field \"relative_url\" on type \"EntityUrl\".",
      "category": "graphql",
      "locations": [
        {
          "line": 4,
          "column": 4
        }
      ]
    }
  ]
}`

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All new or changed code is covered with unit/integration tests (if applicable)
 - [ ] All automated tests passed successfully (all builds on Travis CI are green)

PR #465, opened 2019-03-12
Currency misspelled in graphql attributes

XxXgeoXxX:
## Description (*)
Issue https://github.com/magento/graphql-ce/issues/464

### DIFF
--- a/app/code/Magento/DirectoryGraphQl/etc/schema.graphqls
+++ b/app/code/Magento/DirectoryGraphQl/etc/schema.graphqls
@@ -13,7 +13,7 @@ type Currency {
     default_display_currecy_code: String @deprecated(reason: "Symbol was missed. Use `default_display_currency_code`.")
     default_display_currency_code: String
     default_display_currecy_symbol: String @deprecated(reason: "Symbol was missed. Use `default_display_currency_symbol`.")
-    default_display_currency_symbol
+    default_display_currency_symbol: String
     available_currency_codes: [String]
     exchange_rates: [ExchangeRate]
 }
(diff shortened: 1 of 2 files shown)


TomashKhamlai:
[Here](https://github.com/magento/graphql-ce/pull/398#discussion_r262603127) @naydav requested to keep both values because of BIC (Backward Incompatible Change). Read more about BIC because this could be also requested here. But don't push any commits while not requested. It's just information, not request.

[backward-incompatible-changes](https://devdocs.magento.com/guides/v2.3/release-notes/backward-incompatible-changes/)
[backward-compatible-development](https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/)

app/code/Magento/DirectoryGraphQl/etc/schema.graphqls
```diff
@@ -10,8 +10,8 @@ type Query {
 type Currency {
     base_currency_code: String
     base_currency_symbol: String
-    default_display_currecy_code: String
```
naydav, line 13: It will Backward Incompatible Changes
We need to mark old fields as `deprecated` instead of removing them
thanks

naydav (changes requested):

app/code/Magento/DirectoryGraphQl/etc/schema.graphqls
XxXgeoXxX: Fields were marked as deprecated

app/code/Magento/DirectoryGraphQl/etc/schema.graphqls
```diff
@@ -10,8 +10,10 @@ type Query {
...
-    default_display_currecy_symbol: String
+    default_display_currecy_code: String @deprecated(reason: "Symbol was missed. Use `default_display_currency_code`.")
```
naydav, line 13: Looks like we can use names like `code`, `default_code` etc
It's part of a single object so we don't need to duplicate object name in object fields
BUT old fields should not be removed, just marked as deprecated

naydav (changes requested):

Issue #468, opened 2019-03-12
Resolve coupling between objects in `\Magento\QuoteGraphQl\Model\Cart\SetBillingAddressOnCart`


naydav:
Docblock of `\Magento\QuoteGraphQl\Model\Cart\SetBillingAddressOnCart:execute`
should be
```
/**
     * Set billing address for a specified shopping cart
     *
     * @param ContextInterface $context
     * @param CartInterface $cart
     * @param array $billingAddressInput
     * @return void
     * @throws GraphQlInputException
     * @throws GraphQlAuthenticationException
     * @throws GraphQlAuthorizationException
     * @throws GraphQlNoSuchEntityException
     */
    public function execute(ContextInterface $context, CartInterface $cart, array $billingAddressInput): void
```
But in this case we get static tests error
```
Magento\Test\Php\LiveCodeTest::testCodeMess
PHP Code Mess has found error(s):
/var/www/html/app/code/Magento/QuoteGraphQl/Model/Cart/SetBillingAddressOnCart.php:22	The class SetBillingAddressOnCart has a coupling between objects value of 13. Consider to reduce the number of dependencies under 13.
```

So, need to split logic of ``\Magento\QuoteGraphQl\Model\Cart\SetBillingAddressOnCart`` in several classes

vasilii-b:
Hi @naydav, I'll take a look into this one. Thank you!

naydav:
https://github.com/magento/graphql-ce/pull/488

PR #488, opened 2019-03-16
Resolve coupling between objects in `\Magento\QuoteGraphQl\Model\Cart\SetBillingAddressOnCart`

vasilii-b:
### Description (*)
Fixed issue magento/graphql-ce#468

1.  magento/graphql-ce#468

1. Go to `/dev/tests/static/` (CLI `cd /dev/tests/static/` from project root folder)
2. Prepare your `phpunit.xml` from `phpunit.xml.dist`
3. run `../../../vendor/bin/phpunit`
4. Tests results should be green

image


app/code/Magento/QuoteGraphQl/Model/Cart/CreateQuoteAddressByCustomerAddress.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->getCustomer = $getCustomer;
+        $this->getCustomerAddress = $getCustomerAddress;
+    }
+
+    /**
+     * @param ContextInterface $context
+     * @param int|string|null $customerAddressId
+     * @param array|null $customerAddress
+     * @return Address
+     * @throws GraphQlAuthenticationException
```
TomashKhamlai: Usually `@throws` annotation tags are added only when exception is thrown explicitly in the function. Any other exceptions from parent methods are implicit and not annotated. "Missing `@throws` tag(s)" can be turned off in PHPStorm Inspections->PHP->PHPDoc

TomashKhamlai:
@naydav. Checked. Works as expected

app/code/Magento/QuoteGraphQl/Model/Cart/CreateQuoteAddressByCustomerAddress.php
vasilii-b: Thanks for mentioning, @TomashKhamlai. I've added requested changes. Thank you!

naydav: > Usually @throws annotation tags are added only when exception is thrown explicitly in the function

I little disagree with this statement. If I work with some API I will prefer to know which of exception could be thrown without submerging in implementation (does not matter will be thrown in current method or in some child classes)

Generally, it's about readability.

Issue #71, opened 2018-05-23
Decouple from Webonyx


misha-kotov:
As a Magento Developer, I want to be not depended on the Webonyx library for further iterations of GraphQL Code. 

Currently there are classes in the Magento GraphQL framework (ex: ResolveInfo) that are inherited from Webonyx components. This creates a dependency for Magento and 3rd party developers on Webonyx library. This library has not seen many updates lately and is technically not in a stable version (and may never get there). So the value in decoupling is that we can continue evolving our GraphQL capabilities without being subjected to changes in the Webonyx library itself.  Another benefit is being able to switch to another library in the future if a better one becomes available.

PR #490, opened 2019-03-16
New field resolver architecture proposal

rogyar:
### Description (*)
The general idea of the current PR is starting decoupling from Webonyx.
Currently, within a scope of a resolver, we work with data provided by Webonyx directly. For example:
https://github.com/magento/graphql-ce/blob/7b1cd79738b2cff7881a8defe4a58c6142c8ed88/app/code/Magento/QuoteGraphQl/Model/Resolver/Cart.php#L38-L43

In order to start the decoupling process we need to get rid of the direct usage of the Webonyx data and introduce an intermediate layer (interface) with the corresponding set of getters/setters. At the very beginning the mentioned intermediate layer will work with Webonyx. However, in the future we can change the interface implementation to our own if necessary.

1. #71 : Decouple from Webonyx


lib/internal/Magento/Framework/GraphQl/FieldResolverContextInterface.php
```diff
@@ -0,0 +1,26 @@
...
+declare(strict_types=1);
+
+namespace Magento\Framework\GraphQl;
+
+/**
+ * Interface FieldResolverContextInterface
+ *
+ * Field resolver context
+ */
+interface FieldResolverContextInterface
```
naydav: Please, rename to `QueryContextInterface`

lib/internal/Magento/Framework/GraphQl/Query/QueryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+declare(strict_types=1);
+
+namespace Magento\Framework\GraphQl\Query;
+
+/**
+ * Interface QueryInterface
+ *
+ * Contains information about GraphQl query
+ */
+interface QueryInterface
```
naydav, line 15: Please finish description this and the other of Interface (including parameters and return values)

lib/internal/Magento/Framework/GraphQl/Schema/Type/Output/ElementMapper/Formatter/Fields.php
```diff
@@ -146,6 +146,10 @@ private function getFieldConfig(
             /** @var ResolverInterface $resolver */
             $resolver = $this->objectManager->get($field->getResolver());
 
+            // TODO: Here the new resolver interface will be substituted.
```
naydav: Please, implement the approach for any resolver from `QuoteGraphQl` module

lib/internal/Magento/Framework/GraphQl/Query/QueryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * TBP
+     */
+    public function getStructure();
+
+    /**
+     * TBP
+     */
+    public function getArgs();
```
naydav: Rename to `getArguments`

naydav (changes requested):

rogyar:
Hi @naydav. I've implemented Cart resolver according to the new resolver structure. 
I have doubts about the necessity to introduce a new context interface/implementation since we already have one `\Magento\Framework\GraphQl\Query\Resolver\ContextInterface`. I've reused this interface in my example. However, if you have another point on this matter - I'm open for discussions.

Thank you!

naydav:
We are going to create a new more efficient way for filtration and loading data.
Creation of Design Document is in progress (#567).

So, need to wait to introduce a new mechanism and just after that continue work with this PR.

Issue #480, opened 2019-03-15
[Test Coverage] 'GetBillingAddress' functionality


naydav:
Need to cover `GetBillingAddress` functionality with API-functional tests.

Query scheme: `app/code/Magento/QuoteGraphQl/etc/schema.graphqls`

Testcases should be created::
```
\Magento\GraphQl\Quote\Guest\GetBillingAddressTest
- testGetCartWithBillingAddress
- testGetBillingAddressFromCustomerCart
- testGetBillingAddressIfBillingAddressIsNotSet
- testGetBillingAddressOfNonExistentCart
```
Example: `\Magento\GraphQl\Quote\Guest\GetAvailablePaymentMethodsTest`

```
\Magento\GraphQl\Quote\Customer\GetBillingAddressTest
- testGetCartWithBillingAddress
- testGetBillingAddressFromGuestCart
- testGetBillingAddressFromAnotherCustomerCart
- testGetBillingAddressIfBillingAddressIsNotSet
- testGetBillingAddressOfNonExistentCart
```
Example: `\Magento\GraphQl\Quote\Customer\GetAvailablePaymentMethodsTest`

Please, reuse current fixtures instead of creating new.

More examples you could find in
https://github.com/magento/graphql-ce/tree/2.3-develop/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Customer

https://github.com/magento/graphql-ce/tree/2.3-develop/dev/tests/api-functional/testsuite/Magento/GraphQl/Quote/Guest

atwixfirster:
**Pull request:** https://github.com/magento/graphql-ce/pull/494

Thank you!

naydav:
https://github.com/magento/graphql-ce/pull/494

PR #494, opened 2019-03-16
magento/graphql-ce#480: [Test Coverage] 'GetBillingAddress' functiona…

atwixfirster:
**480 - [Test Coverage] 'GetBillingAddress' functionality**

Issue: https://github.com/magento/graphql-ce/issues/480

Testcases should be created::

\Magento\GraphQl\Quote\Guest\GetBillingAddressTest
- testGetCartWithBillingAddress
- testGetBillingAddressFromCustomerCart
- testGetBillingAddressIfBillingAddressIsNotSet
- testGetBillingAddressOfNonExistentCart

Example: \Magento\GraphQl\Quote\Guest\GetAvailablePaymentMethodsTest

\Magento\GraphQl\Quote\Customer\GetBillingAddressTest
- testGetCartWithBillingAddress
- testGetBillingAddressFromAnotherCustomerCart
- testGetBillingAddressIfBillingAddressIsNotSet
- testGetBillingAddressOfNonExistentCart


dev/tests/integration/testsuite/Magento/Sales/_files/guest_quote_with_addresses.php
```diff
@@ -10,27 +10,31 @@
...
-    ->setStockData(
-        [
-            'qty' => 100,
-            'is_in_stock' => 1,
-        ]
-    )->save();
-
 $productRepository = $objectManager->create(\Magento\Catalog\Api\ProductRepositoryInterface::class);
+
+try {
```
TomashKhamlai: Don't use try-catch construction in fixtures. Every fixture should have a rollback and that's enough.
Short story: reinstall Magento, remove try-catch and run test again it is passing locally.

Long story: Sometimes(very often for me) when test exited with error people try to fix the problem and then they try again. On this step if they do not succeed they will have 2 equal records in the database. But rollback expects only one record... so don't use try-catch

atwixfirster: > Don't use try-catch construction in fixtures. Every fixture should have a rollback and that's enough.
> Short story: reinstall Magento, remove try-catch and run test again it is passing locally.
> 
> Long story: Sometimes(very often for me) when test exited with error people try to fix the problem and then they try again. On this step if they do not succeed they will have 2 equal records in the database. But rollback expects only one record... so don't use try-catch

Thank you for notice, @TomashKhamlai !

Working on solution...

dev/tests/integration/testsuite/Magento/Sales/_files/guest_quote_with_addresses.php
```diff
@@ -10,12 +10,15 @@
 \Magento\TestFramework\Helper\Bootstrap::getInstance()->loadArea(\Magento\Framework\App\Area::AREA_FRONTEND);
 
 $objectManager = \Magento\TestFramework\Helper\Bootstrap::getObjectManager();
+$productRepository = $objectManager->create(\Magento\Catalog\Api\ProductRepositoryInterface::class);
```
naydav: Please, use `get` for stateless services

atwixfirster: > Please, use get for stateless services
yes, I use 
```
$product = $productRepository->get('simple-product-guest-quote');
```
in line 34.

naydav: I mean need to "GET" repository instead of creating it.

atwixfirster: > I mean need to "GET" repository instead of creating it.

**fixed**

Thank you, @naydav !

PR #495, opened 2019-03-16
magento/graphql-ce#: Add missed rollback

atwixfirster:
Added missed rollback for `Magento/Customer/_files/three_customers.php` fixture

### DIFF
--- a/dev/tests/integration/testsuite/Magento/Customer/_files/three_customers_rollback.php
+++ b/dev/tests/integration/testsuite/Magento/Customer/_files/three_customers_rollback.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+use Magento\Integration\Model\Oauth\Token\RequestThrottler;
+
+$objectManager = \Magento\TestFramework\Helper\Bootstrap::getObjectManager();
+
+/** @var \Magento\Framework\Registry $registry */
+$registry = $objectManager->get(\Magento\Framework\Registry::class);
+$registry->unregister('isSecureArea');
+$registry->register('isSecureArea', true);
+
+$customersToRemove = [
+    'customer@search.example.com',
+    'customer2@search.example.com',
+    'customer3@search.example.com',
+];
+
+/**
+ * @var Magento\Customer\Api\CustomerRepositoryInterface
+ */
+$customerRepository = $objectManager->create(\Magento\Customer\Api\CustomerRepositoryInterface::class);
+/**
+ * @var RequestThrottler $throttler
+ */
+$throttler = \Magento\TestFramework\Helper\Bootstrap::getObjectManager()->create(RequestThrottler::class);
+foreach ($customersToRemove as $customerEmail) {
+    try {
+        $customer = $customerRepository->get($customerEmail);
+        $customerRepository->delete($customer);
+    } catch (\Magento\Framework\Exception\NoSuchEntityException $exception) {
+        // not found
+    }
+
+    /* Unlock account if it was locked for tokens retrieval */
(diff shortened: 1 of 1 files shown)


atwixfirster:
**Issue:** https://github.com/magento/graphql-ce/issues/502

dev/tests/integration/testsuite/Magento/Customer/_files/three_customers_rollback.php
```diff
@@ -0,0 +1,43 @@
+<?php
```
naydav, line 1: Please looks like I do it for `two_customers_rollback.php`

naydav (changes requested):

naydav:
Also, please create the corresponding ticket
All of PRs should be linked with ticket
Thanks

atwixfirster:
> Also, please create the corresponding ticket
> All of PRs should be linked with ticket
> Thanks

@naydav , ticket is already created. 😊   I mentioned it in the comment above. 

*Issue:* https://github.com/magento/graphql-ce/issues/502

Thank you!

dev/tests/integration/testsuite/Magento/Customer/_files/three_customers_rollback.php
atwixfirster: @naydav , I do not like `deleteById` solution here:
```php
$customerRepository->deleteById(2);
```
I will explain why.

If you take a look to `Magento\GraphQl\Quote\Customer\GetAvailablePaymentMethodsTest`::`testGetPaymentMethodsFromAnotherCustomerCart()` you will see that for this test we use the next fixtures:
- Magento/Checkout/_files/quote_with_items_saved.php
- Magento/Customer/_files/three_customers.php

Fixture `quote_with_items_saved.php` uses `Customer/_files/customer.php` which creates a customer with `entity_id = 1` and `email = customer@example.com`.

In this case `three_customers.php` is unable to create `customer@search.example.com` with `entity_id=1` because customer with such entity_id is already exist. 

Please see: https://github.com/magento/graphql-ce/blob/2.3-develop/dev/tests/integration/testsuite/Magento/Customer/_files/three_customers.php

So, that is why I decided to delete customers by email in the fixture rollback.

**My opinion:** we should not set **entity_id** for the customer fixtures. If you agree then this is a stuff for the new ticket.

Please review and provide your suggestions. 

Thank you!

naydav: @atwixfirster 
You can't use 
`Magento/Customer/_files/customers.php`
`Magento/Customer/_files/three_customers.php`
at the same time because both of fixture contains hardcoded `customer_id` thus this fix will not resolve the described problem.

BUT, 
1) We can not change current fixtures due to Backward Compatibility
https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/#magento-functional-and-integration-tests
2) you should not use both of these fixtures at the same time. They are designed for different scenarios.

atwixfirster: Understood , @naydav 👍 

I've made a rollback for changes in the `Magento/Customer/_files/customers.php` and left `Magento/Customer/_files/three_customers.php` only.

Thank you again for the explanations!

atwixfirster: @naydav a quick notice for you.

It's not a good point to use both fixtures for the same test:
- Magento/Customer/_files/customers.php
- Magento/Customer/_files/three_customers.php

However , when we use 
- Magento/Customer/_files/three_customers.php
- Magento/Checkout/_files/quote_with_items_saved.php

then `quote_with_items_saved.php` includes `customers.php`.

Issue #477, opened 2019-03-15
In some cases only data is received even when error or validation message is also expected


TomashKhamlai:
### Preconditions (*)

1.
2.

### Steps to reproduce (*)

1. Execute query:
```graphql
mutation {
  addDownloadableProductsToCart(
    input: {
      cart_id: "wwuiBJ2ZuJmwZcYbNBU7bqeX8Y0JDpql"
      cartItems: {
        data: {
          qty: 1
          sku: "downloadable-product-purchase-separately"
        }
        downloadable_product_links: [
          {
            link_id: 9
          }
          {
            link_id: 10
          }
        ]
      }
    }
  ) {
    cart {
      items {
        id
        qty
      }
    }
  }
}
```
2. Execute query:
```graphql
mutation {
  addDownloadableProductsToCart(
    input: {
      cart_id: "wwuiBJ2ZuJmwZcYbNBU7bqeX8Y0JDpql"
      cartItems: {
        data: {
          qty: 0.5
          sku: "downloadable-product-purchase-separately"
        }
        downloadable_product_links: [
          {
            link_id: 9
          }
          {
            link_id: 10
          }
        ]
      }
    }
  ) {
    cart {
      items {
        id
        qty
      }
    }
  }
}
```

### Expected result (*)

1. Validation with data:
```graphql
{
  "errors": [
    {
      "message": "Could not add the product with SKU downloadable-product-purchase-separately to the shopping cart: The fewest you may purchase is 1.",
      "category": "graphql-input",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "addDownloadableProductsToCart"
      ]
    }
  ],
  "data": {
    "addDownloadableProductsToCart": {
      "cart": {
        "items": [
          {
            "id": "11",
            "qty": 1
          }
        ]
      }
    }
  }
}
```

### Actual result (*)

1. Only data:
```graphql
{
  "data": {
    "addDownloadableProductsToCart": {
      "cart": {
        "items": [
          {
            "id": "11",
            "qty": 1
          }
        ]
      }
    }
  }
}
```

atwixfirster:
**Pull request:** https://github.com/magento/graphql-ce/pull/523

atwixfirster:
@TomashKhamlai , I am not sure if this issue is an error.

In the first your query you added to Cart `1` x `Product A`.

In the second one, you added `0.5` x `Product A` to the same shopping cart. 

[...]

PR #523, opened 2019-03-22
magento/graphql-ce#477: In some cases only data is received even when…

atwixfirster:
**477 - In some cases only data is received even when error or validation message is also expected**

Fix for issue: https://github.com/magento/graphql-ce/issues/477

### DIFF
--- a/app/code/Magento/CatalogInventory/Model/Quote/Item/QuantityValidator/Initializer/StockItem.php
+++ b/app/code/Magento/CatalogInventory/Model/Quote/Item/QuantityValidator/Initializer/StockItem.php
@@ -90,11 +90,13 @@ class StockItem
 
         $stockItem->setProductName($product->getName());
 
+        $addToCartQty = $quoteItem->getQtyToAdd();
+
         $result = $this->stockState->checkQuoteItemQty(
             $product->getId(),
             $qtyForCheck,
             $rowQty,
-            ($quoteItem->getQtyToAdd() ? $quoteItem->getQty() - $quoteItem->getQtyToAdd() : $quoteItem->getQty()),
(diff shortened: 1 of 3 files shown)


app/code/Magento/CatalogInventory/Model/Quote/Item/QuantityValidator/Initializer/StockItem.php
```diff
@@ -92,9 +92,9 @@ public function initialize(
 
         $result = $this->stockState->checkQuoteItemQty(
             $product->getId(),
-            $rowQty,
             $qtyForCheck,
-            $qty,
+            $qty,         
+            (is_object($quoteItem->getStockStateResult()) ? $quoteItem->getStockStateResult()->getOrigQty() : $quoteItem->getQty()),
```
naydav, line 97: @maghamed Could you provide your feedback?
This part looks very strange

atwixfirster: I am not sure that created https://github.com/magento/graphql-ce/issues/477 issue is a real issue.

I described a situation in the issue ticket . However, I can add explanations here:

In the first Tomash query he added to Cart `1` x `Product A`.

In the second one, Tomash added `0.5` x `Product A` to the same shopping cart. 

So, in the background process Tomash added `1.5` x `Product A` into shopping cart because `Product A` with quantity `1` already exists in the cart + a new quantity `0.5`. 

In the result, Tomash added to cart `1.5` x `Product A`. 

`Product A` does not support decimal product qty that is why `1.5` has been converted into `1`.

Thank you!

naydav:
Closed due to https://github.com/magento/graphql-ce/issues/477#issuecomment-486463399

Issue #524, opened 2019-03-22
Cannot return null for non-nullable field AvailableShippingMethod.method_code when no shipping methods are available


TomashKhamlai:
### Preconditions (*)

1. Create Customer with US address (default billing and default shipping address)

### Steps to reproduce (*)

1. Navigate to Stores> Settings>Configuration>Sales>Shipping Methods and turn off Flat Rate
2. Enable Free Shipping but go to **Ship to Applicable Countries** and uncheck **Use Default**
3. In **Ship to Specific Countries** set country to United Kingdom only.
4. In **Show Method if Not Applicable** set Yes
5. Get Customer Token
6. Create empty Cart
7. Add simple product to cart
8. Execute query `cart`:
```graphql

query {
  cart(
    cart_id: "mfdHyHY0QgtXW5LeHxfJ0nVCjWjy655n"
  ) {
    available_payment_methods {
      code
      title
    }
    shipping_addresses {
      available_shipping_methods {
        carrier_title
        method_code
      }
    }
    
  }
}
```

### Expected result (*)

```
"available_shipping_methods": [
            {
              "carrier_code": "freeshipping",
              "carrier_title": "Free Shipping",
              "method_code": null,
              "method_title": null,
              "error_message": "This shipping method is not available. To use this shipping method, please contact us.",
              "amount": 0,
              "base_amount": null,
              "price_excl_tax": 0,
              "price_incl_tax": 0,
              "available": false
            }
```

### Actual result (*)

1. Internal Server Error:
```graphql
{
  "errors": [
    {
      "debugMessage": "Cannot return null for non-nullable field AvailableShippingMethod.method_code.",
      "message": "Internal server error",
      "category": "internal",
      "locations": [
        {
          "line": 10,
          "column": 9
        }
      ],
      "path": [
        "cart",
        "shipping_addresses",
        0,
        "available_shipping_methods",
        0,
        "method_code"
      ]
    }
  ],
  "data": {
    "cart": {
      "available_payment_methods": [
        {
          "code": "checkmo",
          "title": "Check / Money order"
        }
      ],
      "shipping_addresses": [
        {
          "available_shipping_methods": [
            null
          ]
        }
      ]
    }
  }
}
```

naydav:
https://github.com/magento/graphql-ce/pull/526

PR #526, opened 2019-03-22
Cannot return null for non-nullable field AvailableShippingMethod.method_code when no shipping methods are available

XxXgeoXxX:
### Description (*)
Description https://github.com/magento/graphql-ce/issues/524

Steps to reproduce (*)
Navigate to Stores> Settings>Configuration>Sales>Shipping Methods and turn off Flat Rate
Enable Free Shipping but go to Ship to Applicable Countries and uncheck Use Default
In Ship to Specific Countries set country to United Kingdom only.
In Show Method if Not Applicable set Yes
Get Customer Token
Create empty Cart
Add simple product to cart
Execute query cart:
```
query {
  cart(
    cart_id: "mfdHyHY0QgtXW5LeHxfJ0nVCjWjy655n"
  ) {
    available_payment_methods {
      code
      title
    }
    shipping_addresses {
      available_shipping_methods {
        carrier_title
        method_code
      }
    }
    
  }
}
```


app/code/Magento/QuoteGraphQl/Model/Resolver/ShippingAddress/AvailableShippingMethods.php
```diff
@@ -65,13 +66,19 @@ public function resolve(Field $field, $context, ResolveInfo $info, array $value
...
                     [],
                     ShippingMethodInterface::class
                 );
+                if ($method['available'] && $method['error_message'] === "") {
+                    $methods[] = $method;
+                }
             }
         }
+        if (count($methods) === 0) {
+            throw new GraphQlNoSuchEntityException(__(' This shipping method is not available. To use this shipping method, please contact us.'));
```
TomashKhamlai: Is it possible to load the text of the message from config [there](
https://github.com/magento/graphql-ce/blob/7b885666e9115441fe136202c60dc41044311b2f/app/code/Magento/OfflineShipping/Model/Carrier/Freeshipping.php#L104)

Have no idea how it can be done. So it's a question, not suggestion.

app/code/Magento/QuoteGraphQl/Model/Resolver/ShippingAddress/AvailableShippingMethods.php
```diff
@@ -65,13 +66,19 @@ public function resolve(Field $field, $context, ResolveInfo $info, array $value
         $shippingRates = $address->getGroupedAllShippingRates();
```
naydav: Need to check this behavior on frontend.
How to frontend work with `getGroupedAllShippingRates` method and provide the same error message

naydav (changes requested):

galaoleksandr:
Cannot reproduce the issue.

XxXgeoXxX:
Hi, @galaoleksandr
This bug reproduced with enabled debugging.
Browser extension return empty array instead shipping error message

naydav:
@TomashKhamlai Why do you consider that we need to show an error message?

Let's go step by step:
1) Create cart
2) Adding product to cart
3) Get info about the cart
```
query {
  cart(
    cart_id: "mfdHyHY0QgtXW5LeHxfJ0nVCjWjy655n"
  ) {
    available_payment_methods {
      code
      title
    }
    shipping_addresses {
      available_shipping_methods {
        carrier_title
        method_code
      }
    }    
  }
}
```

Result should be
```
{
  "data": {
    "cart": {
      "available_payment_methods": [
        {
          "code": "checkmo",
          "title": "Check / Money order"
        }
      ],
      "shipping_addresses": [
        {
          "available_shipping_methods": []
        }
      ]
    }
  }
}
```
So, I see that there are not any `available shipping methods`.
This is enough information about deciding decision on UI.

@paliarush Could you look at it?

naydav:
@XxXgeoXxX 
In the scope of this task need to make 
```
type AvailableShippingMethod {
    carrier_code: String!
    carrier_title: String!
    method_code: String!
    method_title: String!
    error_message: String
    amount: Float!
    base_amount: Float!
    price_excl_tax: Float!
    price_incl_tax: Float!
}
```
without required fields because it is not `InputType`(which can has some validation rules for input parameters) but it is `Type`
So we don't need to force to load all information (moreover GraphQL is way how to load just what you need)

TomashKhamlai:
@naydav, the message appears on Storefront and it also appears in developer mode.

Issue #387, opened 2019-02-22
Test coverage of getting IDs of CMS page/blocks by GraphQL API


galaoleksandr:
Cover getting ID of CMS page and IDs of CMS blocks with API-functional tests.

atwixfirster:
**PR:** https://github.com/magento/graphql-ce/pull/585

Thank you!

naydav:
@galaoleksandr @atwixfirster 
Looks like we don't need to expand scheme with surrogate(internal/technical) keys.

`page_id` for `Page` and `block_id` for Block are surrogate keys, you should use `identifier` for loading of these entities.

The same situation with a Product where `sku` is a natural identifier but not `entity_id` (needed just for internal purposes)
That's why we don't provide a possibility to get Product by `entity_id` in GraphQL

But we still have these fields in Api
`\Magento\Cms\Api\Data\BlockInterface::BLOCK_ID`
`\Magento\Cms\Api\Data\PageInterface::PAGE_ID`

@paliarush 
Please, share you opinion
Thanks

naydav:
https://github.com/magento/graphql-ce/pull/585

PR #585, opened 2019-04-10
387-Test coverage of getting IDs of CMS page/blocks by GraphQL API

atwixfirster:
**Issue:** https://github.com/magento/graphql-ce/issues/387

Thank you!


app/code/Magento/CmsGraphQl/etc/schema.graphqls
```diff
@@ -13,13 +13,15 @@ type StoreConfig @doc(description: "The type contains information about a store
 type Query {
     cmsPage (
         id: Int @doc(description: "Id of the CMS page")
+        identifier: String @doc(description: "Identifier of the CMS page")
```
vpodorozh, line 16: You have added 'identifier' in the query but not added 'identifier' in CmsPage type. Why?
I think it would be logical to get an identifier in a return result as well.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -44,10 +62,32 @@ public function __construct(
      * @return array
      * @throws NoSuchEntityException
      */
-    public function getData(int $pageId): array
```
vpodorozh, line 49: Public method was removed - it is forbidden.
Check this article: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/
**section** "Public and protected method removal"
Please mark `getData` method as a _@depricated_ instead of removing it.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -44,10 +62,32 @@ public function __construct(
...
         $page = $this->pageRepository->getById($pageId);
 
+        return $this->convertPageData($page);
+    }
+
+    /**
+     * @param string $pageIdentifier
+     * @return array
+     */
+    public function getDataByPageIdentifier(string $pageIdentifier): array
```
vpodorozh, line 82: You have added new method `getDataByPageIdentifier`, this should be done in a different way - please check this article: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/
**section** "Introduction of a method to a class or interface"

I think you can solve "public method removal" and "adding a new public method" issues using this approach:
- create two separate handlers for getting data by id `PageById` and by identifier `PageByIdentifier`.
- mark '\Magento\CmsGraphQl\Model\Resolver\DataProvider\Page:getData' method as a deprecated and inject their usage of `PageById` instead of code duplication.
- In `\Magento\CmsGraphQl\Model\Resolver\Page` should be injected `PageById` and `PageByIdentifier` for getting CmsPage data by page id and identifier.
In this case, you will follow the backward compatibility guides.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -44,10 +62,32 @@ public function __construct(
      * @return array
      * @throws NoSuchEntityException
      */
-    public function getData(int $pageId): array
+    public function getDataByPageId(int $pageId): array
```
vpodorozh, line 68: You have added new method `getDataByPageId`, this should be done in a different way - please check this article: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/
**section** "Introduction of a method to a class or interface"
Check my prev comment for solution description.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -18,24 +20,40 @@
...
+     */
+    private $widgetFilter;
+
+    /**
+     * @param GetPageByIdentifierInterface $getPageByIdentifier
      * @param FilterEmulate $widgetFilter
+     * @param PageRepositoryInterface $pageRepository
+     * @param StoreManagerInterface $storeManager
      */
     public function __construct(
```
vpodorozh, line 48: You have changed `\Magento\CmsGraphQl\Model\Resolver\DataProvider\Page::__construct` params not according to `backward-compatible-development` guide.
Please check the correct way in this article: https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/
**section** "Adding a constructor parameter"

vpodorozh (changes requested):
@atwixfirster - thank you for your contribution!
Please review my change request - it is mostly related to backward compatibility.
Thx!

app/code/Magento/CmsGraphQl/etc/schema.graphqls
atwixfirster: > You have added 'identifier' in the query but not added 'identifier' in CmsPage type. Why?

forgot 😊 

> I think it would be logical to get an identifier in a return result as well.

**fixed**

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
atwixfirster: I've created a new data provider model instead of make changes in the current one. 

Could you please review?

Thank you, @vpodorozh !

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
atwixfirster: I've created a new data provider model - `app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/PageDataProvider.php` instead of make changes in the current one - `app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php`.

Please review a new solution.

Thank you!

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
atwixfirster: **fixed**

Thank you!

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
atwixfirster: **fixed**

Thank you!

naydav:
@galaoleksandr @atwixfirster
Looks like we don't need to expand scheme with surrogate(internal/technical) keys.

`page_id` for `Page` and `block_id` for `Block` are surrogate keys, you should use `identifier` key for loading of these entities.

The same situation with a `Product` where `sku` is a natural identifier but not `entity_id` (needed just for internal purposes)
That's why we don't provide a possibility to get `Product` by `entity_id` in GraphQL

But we still have these fields in Api
`\Magento\Cms\Api\Data\BlockInterface::BLOCK_ID`
`\Magento\Cms\Api\Data\PageInterface::PAGE_ID`

@paliarush
Please, share your opinion
Thanks

vpodorozh:
Hi @naydav - thx for clarifications.
I want to make sure that I've understood you correctly, so we need to do the next:
1. Do not expand `type CmsBlock` with `block_id` - `identifier` should be used instead;
2. `type CmsPage` should be expanded only with `identifier` (`page_id` should not be added);
3. Query interface for `cmsPage` should be expanded with `identifier` argument - to make it possible to get data via this key (but not `id`);
4. Query interface for `cmsPage` should keep `id` argument to keep backwards-compatibility;

Am I right?

naydav:
> Hi @naydav - thx for clarifications.
> I want to make sure that I've understood you correctly, so we need to do the next:
> 
> 1. Do not expand `type CmsBlock` with `block_id` - `identifier` should be used instead;
> 2. `type CmsPage` should be expanded only with `identifier` (`page_id` should not be added);
> 3. Query interface for `cmsPage` should be expanded with `identifier` argument - to make it possible to get data via this key (but not `id`);
> 4. Query interface for `cmsPage` should keep `id` argument to keep backwards-compatibility;
> 
> Am I right?

Correct (and mark it as deprecated with a resolution like use `use identifier`)
Thanks

app/code/Magento/CmsGraphQl/etc/schema.graphqls
```diff
@@ -13,13 +13,16 @@ type StoreConfig @doc(description: "The type contains information about a store
...
     ): CmsPage @resolver(class: "Magento\\CmsGraphQl\\Model\\Resolver\\Page") @doc(description: "The CMS page query returns information about a CMS page")
     cmsBlocks (
         identifiers: [String] @doc(description: "Identifiers of the CMS blocks")
     ): CmsBlocks @resolver(class: "Magento\\CmsGraphQl\\Model\\Resolver\\Blocks") @doc(description: "The CMS block query returns information about CMS blocks")
 }
 
 type CmsPage @doc(description: "CMS page defines all CMS page information") {
+    page_id: Int @doc(description: "Entity ID of CMS page")
+    identifier: String @doc(description: "Identifier of the CMS page")
```
vpodorozh, line 24: Hi @atwixfirster - please check comment from @naydav about GraphQL scheme and apply needed changes.
Thx!
https://github.com/magento/graphql-ce/pull/585#issuecomment-484244309

vpodorozh (changes requested):

app/code/Magento/CmsGraphQl/etc/schema.graphqls
atwixfirster: **implemented**

Please review!

Thank you!

atwixfirster: @vpodorozh , could you please pay your attention to this PR?

Thank you!

vpodorozh:
Note, I've reported a separate issue for fixing integration tests: #654 
`1) Magento\GraphQlCache\Controller\Cms\CmsPageCacheTest::testToCheckCmsPageRequestCacheTags
Failed asserting that two arrays are equal.`

Seems that this failed test caused by changes in cache tag policy for cms page in magento core.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -13,6 +13,9 @@
 use Magento\Widget\Model\Template\FilterEmulate;
 
 /**
+ * @deprecated
+ * @see Magento\CmsGraphQl\Model\Resolver\DataProvider\PageDataProvider
```
vpodorozh: Please mark all public methods as deprecated as well.
Check section 'Interface/class removal' in https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/

app/code/Magento/CmsGraphQl/Model/Resolver/Page.php
```diff
@@ -7,7 +7,7 @@
 
 namespace Magento\CmsGraphQl\Model\Resolver;
 
-use Magento\CmsGraphQl\Model\Resolver\DataProvider\Page as PageDataProvider;
+use Magento\CmsGraphQl\Model\Resolver\DataProvider\PageDataProvider as PageDataProvider;
```
vpodorozh: By this, you are changing `$pageDataProvider` argument type in method  `__construct` - this is forbidden according to this article: Check 'Modifying the method argument type' section in https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/  

So you will need to add a new property and argument in `__construct` for your new handler. Do not remove the old one handler argument in `__construct`.

app/code/Magento/CmsGraphQl/etc/schema.graphqls
```diff
@@ -12,14 +12,16 @@ type StoreConfig @doc(description: "The type contains information about a store
 
 type Query {
     cmsPage (
-        id: Int @doc(description: "Id of the CMS page")
+        id: Int @doc(description: "Id of the CMS page") @deprecated(reason: "Use `identifier`") @doc(description: "The CMS page query returns information about a CMS page")
```
vpodorozh: Could you please add a more detailed description for the deprecated tag. Something like: 
>The `id` is deprecated. Use `identifier` instead.

app/code/Magento/CmsGraphQl/Model/Resolver/Page.php
```diff
@@ -44,35 +45,61 @@ public function resolve(
...
      * @param int $pageId
      * @return array
      * @throws GraphQlNoSuchEntityException
      */
-    private function getPageData(int $pageId): array
+    private function getPageDataById(int $pageId): array
+    {
+        try {
+            $pageData = $this->pageDataProvider->getDataByPageId($pageId);
+        } catch (NoSuchEntityException $e) {
```
vpodorozh: Could you please move this `try/catch` statement into `\Magento\CmsGraphQl\Model\Resolver\Page::resolve` and use it for both - `getDataByPageId` and `getPageDataByIdentifier` instead of copy&paste it?
It seems that this method (`\Magento\CmsGraphQl\Model\Resolver\Page::getPageDataById`) will not be needed at all in case you will move `try/catch` statement.

app/code/Magento/CmsGraphQl/Model/Resolver/Page.php
```diff
@@ -44,35 +45,61 @@ public function resolve(
...
+    /**
+     * @param string $pageIdentifier
+     * @return array
+     * @throws GraphQlNoSuchEntityException
+     */
+    private function getPageDataByIdentifier(string $pageIdentifier): array
     {
         try {
-            $pageData = $this->pageDataProvider->getData($pageId);
+            $pageData = $this->pageDataProvider->getDataByPageIdentifier($pageIdentifier);
```
vpodorozh: Could you please move this `try/catch` statement into `\Magento\CmsGraphQl\Model\Resolver\Page::resolve` and use it for both - `getDataByPageId` and `getPageDataByIdentifier` instead of copy&paste it?
It seems that this method (`\Magento\CmsGraphQl\Model\Resolver\Page::getPageDataByIdentifier`) will not be needed at all in case you will move `try/catch` statement.

vpodorozh (changes requested):
@atwixfirster - could you please take a look at my change request?
Thx!

app/code/Magento/CmsGraphQl/Model/Resolver/Page.php
atwixfirster: **done**

app/code/Magento/CmsGraphQl/etc/schema.graphqls
atwixfirster: Description has been extended. 

Thank you for care :)

app/code/Magento/CmsGraphQl/Model/Resolver/Page.php
atwixfirster: **done**

app/code/Magento/CmsGraphQl/Model/Resolver/Page.php
atwixfirster: Thank you!

**implemented**

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
atwixfirster: **done**

vpodorozh:
I'm checking the PR

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
vpodorozh: Okay, so you have changed the solution. Will review in other thread.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -18,28 +20,47 @@
...
+    private $widgetFilter;
+
+    /**
+     * @param GetPageByIdentifierInterface $getPageByIdentifier
      * @param FilterEmulate $widgetFilter
+     * @param PageRepositoryInterface $pageRepository
+     * @param StoreManagerInterface $storeManager
      */
     public function __construct(
+        GetPageByIdentifierInterface $getPageByIdentifier,
```
vpodorozh: 1. \Magento\CmsGraphQl\Model\Resolver\DataProvider\Page::__construct
You have changed the order of original arguments; That is not okay according to backwards-compatibility policy.
2. \Magento\CmsGraphQl\Model\Resolver\DataProvider\Page::__construct
Check "Adding a constructor parameter" section here https://devdocs.magento.com/guides/v2.3/contributor-guide/backward-compatible-development/
That was done not according to backwards-compatibility policy.

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
```diff
@@ -69,4 +90,56 @@ public function getData(int $pageId): array
...
+
+        return $this->convertPageData($page);
+    }
+
+    /**
+     * @param PageInterface $page
+     * @return array
+     * @throws NoSuchEntityException
+     */
+    private function convertPageData(PageInterface $page)
```
vpodorozh, line 97: 3. Reuse the newly created method `convertPageData` in `getData` method - so there will be no code duplication;

vpodorozh (changes requested):
Hi, last circle of review for backward compatibility policy.
Thx!

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
atwixfirster: **done**

Could you please review all changes now?

Thank you!

app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Page.php
atwixfirster: **done**, sir

vpodorozh:
@atwixfirster - I've just got insight for @naydav about backwards-compatibility policy in GraphQl project.
**For graphql - only schema can be counted as a contract.**

So sorry for my snobbism about backwards-compatibility policy :)

vpodorozh (approved):

atwixfirster:
> @atwixfirster - I've just got insight for @naydav about backwards-compatibility policy in GraphQl project.
> **For graphql - only schema can be counted as a contract.**
> 
> So sorry for my snobbism about backwards-compatibility policy :)

sounds like APPROVE? 😊 

Thank you for the reviewing!

Issue #535, opened 2019-03-26
Expose catalog configuration data as a new info query type.


zetlen:
### Description (*)
Exposure of catalog configuration data as a `catalogInfo` query type.

### Expected behavior (*)
The data whose defaults are set in [Magento_Catalog/etc/config.xml](https://github.com/magento/magento2ce/blob/2.3-develop/app/code/Magento/Catalog/etc/config.xml) and whose overrides can be set in the database should be available for query, in order to configure, build, and deploy Progressive Web Apps which honor that configuration.

### Benefits
Unlike most other PWAs, PWA Studio and Venia aim to be as controllable by the Magento 2 admin UI as possible. If the administrator sets a new default for any of these fields, a PWA should do its best to honor that. Most of the time, that requires making a large query to the store instance while rendering the PWA, so things like pagination defaults and URL suffixes can be built into the runtime.

### Additional information

This issue came up when the PWA team examined how to treat URL suffixes for SEO-friendly catalog URLs. Initially, we simply hardcoded the standard `.html` extension to check for pages which we need to dynamically route with a GraphQL query to `urlResolver`. There are two problems with that:

- CMS pages and other custom pages need not have the `.html` extension
- The actual extension is configurable by the administrator in Configuration > Catalog > Search Engine Optimization > Category URL Suffix / Product URL Suffix.

Therefore, we need to be able to access this through some API. If the information is privileged, then we can begin using authentication over GraphQL if necessary and doing the call on the server side alone.

Relevant issues:
https://github.com/magento-research/pwa-studio/pull/936
https://github.com/magento-research/pwa-studio/pull/1029

VitaliyBoyko:
Hi @zetlen 
The functionality to get configs has been implemented within type StoreConfig (basic implementation and usage see here #182).
Could you please define which exact configs you need? So we can add them.
Thank you!

zetlen:
@VitaliyBoyko As we implement features in Venia, we try to use store configuration instead of hardcoding things, so that Magento admins can control their PWA display as much as possible. Therefore, we may eventually use [everything inside the `<catalog />` element](https://github.com/magento/magento2ce/blob/2.3-develop/app/code/Magento/Catalog/etc/config.xml#L10-L55).

If you can make a single change that exposes all of those values, we would like that. None of them are sensitive; they all drive frontend behavior. Could we just put all of them into the `storeConfig` type?

[...]

PR #650, opened 2019-05-01
GraphQl-535: provided catalog configs

VitaliyBoyko:
### Description (*)
Provided catalog configs for PWA in scope of StoreConfig type.

```
{
  storeConfig{
    product_url_suffix,
    category_url_suffix,
    title_separator,
    list_mode,
    grid_per_page_values,
    list_per_page_values,
    grid_per_page,
    list_per_page,
    catalog_default_sort_by
  }
}
```

image

1. #535: Expose catalog configuration data as a new info query type.

### DIFF
--- a/app/code/Magento/CatalogGraphQl/etc/graphql/di.xml
+++ b/app/code/Magento/CatalogGraphQl/etc/graphql/di.xml
@@ -80,4 +80,26 @@
     </virtualType>
     <preference for="Magento\Framework\Search\Adapter\Mysql\Query\Builder\Match"
                 type="Magento\CatalogGraphQl\Model\Search\Adapter\Mysql\Query\Builder\Match" />
+    <type name="Magento\StoreGraphQl\Model\Resolver\Store\StoreConfigDataProvider">
+        <arguments>
+            <argument name="extendedConfigData" xsi:type="array">
+                <item name="product_url_suffix" xsi:type="string">catalog/seo/product_url_suffix</item>
+                <item name="category_url_suffix" xsi:type="string">catalog/seo/category_url_suffix</item>
+                <item name="product_use_categories" xsi:type="string">catalog/seo/product_use_categories</item>
+                <item name="save_rewrites_history" xsi:type="string">catalog/seo/save_rewrites_history</item>
+                <item name="title_separator" xsi:type="string">catalog/seo/title_separator</item>
+                <item name="category_canonical_tag" xsi:type="string">catalog/seo/category_canonical_tag</item>
+                <item name="product_canonical_tag" xsi:type="string">catalog/seo/product_canonical_tag</item>
+                <item name="list_mode" xsi:type="string">catalog/frontend/list_mode</item>
+                <item name="grid_per_page_values" xsi:type="string">catalog/frontend/grid_per_page_values</item>
+                <item name="list_per_page_values" xsi:type="string">catalog/frontend/list_per_page_values</item>
+                <item name="grid_per_page" xsi:type="string">catalog/frontend/grid_per_page</item>
+                <item name="list_per_page" xsi:type="string">catalog/frontend/list_per_page</item>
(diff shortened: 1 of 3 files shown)


app/code/Magento/CatalogGraphQl/etc/graphql/di.xml
```diff
@@ -80,4 +80,26 @@
     </virtualType>
     <preference for="Magento\Framework\Search\Adapter\Mysql\Query\Builder\Match"
                 type="Magento\CatalogGraphQl\Model\Search\Adapter\Mysql\Query\Builder\Match" />
+    <type name="Magento\StoreGraphQl\Model\Resolver\Store\StoreConfigDataProvider">
+        <arguments>
+            <argument name="extendedConfigData" xsi:type="array">
```
naydav, line 85: @VitaliyBoyko Please, check(grep) that all of the items are used and needed on frontend
For example, I have a debt about `flat_catalog_category`...

naydav (changes requested):

app/code/Magento/CatalogGraphQl/etc/graphql/di.xml
VitaliyBoyko: @zetlen Could you please take a look for more configs which might be redundant. Thank you!

naydav: @VitaliyBoyko just grep all of the options in code
If option is not used on frontend (and frontend logic is not based on option) need to delete it from list

naydav (changes requested):

app/code/Magento/CatalogGraphQl/etc/graphql/di.xml
zetlen: I think you're right that we don't need `flat_catalog_category`. Let me list what I can see now will be useful to the PWA frontend in the future. **Anything I _don't_ list here can be removed from this PR if you need to.**

### SEO
- `product_url_suffix`: PWA **definitely** needs this to do accurate SEO routing; these configurations tell our router how to derive the URL key from the URL pathname.
- `category_url_suffix`: PWA **definitely** needs this, see above
- `product_use_categories`: PWA **definitely** needs this, see above
- `title_separator`: PWA can use this to render dynamic page titles.

### Display
- `list_mode`: PWA can use this to control category/search rendering and pagination.
- `grid_per_page_values`: See above
- `list_per_page_values`: See above
- `grid_per_page`: See above
- `list_per_page`: See above
- `catalog_default_sort_by`: See above

Nothing else from the Catalog config section is obviously useful to us (yet). @jimbo and @soumya-ashok and @tjwiebell and @supernova-at might have a different opinion based on UX best practice or known UI gaps, but I think this is plenty.

Thanks again!

zetlen (commented):
I think you're right that we don't need `flat_catalog_category`. Let me list what I can see now will be useful to the PWA frontend in the future. **Anything I _don't_ list here can be removed from this PR if you need to.**

### SEO
- `product_url_suffix`: PWA **definitely** needs this to do accurate SEO routing; these configurations tell our router how to derive the URL key from the URL pathname.
- `category_url_suffix`: PWA **definitely** needs this, see above
- `product_use_categories`: PWA **definitely** needs this, see above
- `title_separator`: PWA can use this to render dynamic page titles.

### Display
- `list_mode`: PWA can use this to control category/search rendering and pagination.
- `grid_per_page_values`: See above
- `list_per_page_values`: See above
- `grid_per_page`: See above
- `list_per_page`: See above
- `catalog_default_sort_by`: See above

Nothing else from the Catalog config section is obviously useful to us (yet). @jimbo and @soumya-ashok and @tjwiebell and @supernova-at might have a different opinion based on UX best practice or known UI gaps, but I think this is plenty.

Thanks again!

naydav:
@VitaliyBoyko, please, update PR according to last comment
Thanks

PR #665, opened 2019-05-03
Prevent Error When Getting Payment Method Before Setting

pmclain:
### Description (*)
Calling [`Magento\Quote\Model\Quote\Payment::getMethodInstance`](https://github.com/magento/graphql-ce/blob/b2ce2a37d921b5ad88fc38663fc0ff3dd6c582d1/app/code/Magento/Payment/Model/Info.php#L105)
throws an exception when the method is not set. This would trigger and error
when requsting the `selected_payment_method` cart property. This commit returns
an empty string instead.

1. magento/graphql-ce#667

1. Create empty cart
```graphql
mutation {
   createEmptyCart
}
```
2. Query cart for selected payment method
```graphql
query getCart($cart_id:String!) {
  cart(cart_id:$cart_id) {
    selected_payment_method {
      code
      title
    }
  }
}
```

### DIFF
--- a/app/code/Magento/QuoteGraphQl/Model/Resolver/SelectedPaymentMethod.php
+++ b/app/code/Magento/QuoteGraphQl/Model/Resolver/SelectedPaymentMethod.php
@@ -34,9 +34,15 @@ class SelectedPaymentMethod implements ResolverInterface
             return [];
         }
 
+        try {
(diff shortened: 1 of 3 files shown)


naydav:
@pmclain Good point!

@TomashKhamlai, please, check the same behavior with shipping methods
thanks

pmclain:
This should resolve #667

Issue #635, opened 2019-04-29
cms block, get by identifier and 


lsv:
### Preconditions (*)

1. Magento 2.3.2

### Steps to reproduce (*)

1. Create atleast 2 store views
2. Create 2 cms blocks, with the same identifier,  set storeview scope two the 2 different store views created at the top
3: Get the cms block with the identifier - it will ALWAYS get the cms block from the default webstore.
4: (You can check with `storeConfig: { code }` that you actually are running with a different scope.

My current setup 

![Screenshot from 2019-04-29 13-45-18](https://user-images.githubusercontent.com/20708/56894088-178fb780-6a85-11e9-9c58-0c67af394cfc.png)

### Expected result (*)

1. The content from the cms block with the content from the scoped cms block

### Actual result (*)

1. Always the default cms block, properly actually the first element found in the database with the cms block identifier.

From my current setup (see screenshot above) when I run the following graphql

```
{
  storeConfig {
    code
  }
  cmsBlocks(identifiers: "footer-block-one") {
    items {
      content
    }
  }
}
```

I get this result

```
{
  "data": {
    "storeConfig": {
      "code": "english"
    },
    "cmsBlocks": {
      "items": [
        {
          "content": "<p>Danish</p>"
        }
      ]
    }
  }
}
```

Where the content should be english

3: If you disable the first cms block (the Danish one in my above setup), you will be thrown a `NotFoundException`, even though the english cms block still is enabled.

```
https://github.com/magento/graphql-ce/blob/2.3-develop/app/code/Magento/CmsGraphQl/Model/Resolver/DataProvider/Block.php#L51
```

A storeview code should be added around here

naydav:
@TomashKhamlai 
Please, verify this issue

GovindaSharma:
#dmcdindia1

GovindaSharma:
i am working on this

TomashKhamlai:
Just a little note:
It is possible to get
```graphql
    "storeConfig": {
      "code": "english"
    }
```
only if the `code` for Default Store View was renamed from "default" to "english" or  "english" is the second Store View and was set as default Store View for the Store. I haven't check the conditions when the main website was removed.

TomashKhamlai:
@lsv could you share your experience of using `storeConfig`?

[...]

PR #669, opened 2019-05-04
cms block, get by identifier and #635

GovindaSharma:
fixed issue #635 cms block, get by identifier

### Description (*)
fixed issue #635 cms block, get by identifier

1. magento/graphql-ce#635: cms block, get by identifier

Steps to reproduce (*)

    Create atleast 2 store views
    Create 2 cms blocks, with the same identifier, set storeview scope two the 2 different store views created at the top
    3: Get the cms block with the identifier - it will ALWAYS get the cms block from the default webstore.
    4: (You can check with storeConfig: { code } that you actually are running with a different scope.

Now the query will be
```graphql
{
  storeConfig {
    code
  }
  cmsBlocks(identifiers: "test_page",
  scope_code: ""
  ) {
    items {
      content
    }
  }
}
```

in scope_code store code need to be passed and it will fetch according to store 

### DIFF
--- a/app/code/Magento/CmsGraphQl/etc/schema.graphqls
+++ b/app/code/Magento/CmsGraphQl/etc/schema.graphqls
@@ -15,7 +15,8 @@ type Query {
         id: Int @doc(description: "Id of the CMS page")
     ): CmsPage @resolver(class: "Magento\\CmsGraphQl\\Model\\Resolver\\Page") @doc(description: "The CMS page query returns information about a CMS page") @cache(cacheTag: "cms_p", cacheIdentity: "Magento\\CmsGraphQl\\Model\\Resolver\\Page\\Identity")
     cmsBlocks (
-        identifiers: [String] @doc(description: "Identifiers of the CMS blocks")
+        identifiers: [String] @doc(description: "Identifiers of the CMS blocks"),
+        scope_code: [String] @doc(description: "Scope Code of the CMS blocks")
     ): CmsBlocks @resolver(class: "Magento\\CmsGraphQl\\Model\\Resolver\\Blocks") @doc(description: "The CMS block query returns information about CMS blocks") @cache(cacheTag: "cms_b", cacheIdentity: "Magento\\CmsGraphQl\\Model\\Resolver\\Block\\Identity")
 }
 
--- a/app/code/Magento/CmsGraphQl/Model/Resolver/Blocks.php
+++ b/app/code/Magento/CmsGraphQl/Model/Resolver/Blocks.php
@@ -46,8 +46,8 @@ class Blocks implements ResolverInterface
     ) {
 
         $blockIdentifiers = $this->getBlockIdentifiers($args);
-        $blocksData = $this->getBlocksData($blockIdentifiers);
-
+        $scopeIdentifiers  = $this->getScopeIdentifiers($args);
+        $blocksData = $this->getBlocksData($blockIdentifiers,$scopeIdentifiers[0]);
         $resultData = [
             'items' => $blocksData,
         ];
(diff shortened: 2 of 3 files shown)


TomashKhamlai:
TestCase 1: (changes required)
getCmsScopeCodeNoSuchEntity
message should be  `"message": "Store View Does Not Exist"`
and probably it is will even better to provide more detail so that we get `"message": "Store View with scope_code \"%scope_code\" Does Not Exist"` but the main idea is not to confuse people with Store/Store View

TomashKhamlai:
Test Case 2: (no changes are required, test coverage only)
Create additional website with store and store view
Create CMS Block for the Main Website on Default Store View that has some text content like: `<p>text{{number of store from StoreManagerInterface}}</p>`
Create CMS Block for the store view on additional website that has some text content like: `<p>text{{number of store from StoreManagerInterface}}</p>`
Assert the content of the block from the first store.
Assert the content of the block from the second store.
Assert that contents are different

TomashKhamlai:
@naydav, I will create few issues found during testing this PR as a separate tasks.
@GovindaSharma, we encourage people to cover pull requests with api-functional tests. You are the author of the PR and this right primarily belongs to you. But if it is too complicated for you, please create an issue like it was done [here](https://github.com/magento/graphql-ce/issues/387)

TomashKhamlai:
#671 reproducible here

TomashKhamlai:
@GovindaSharma you have uncompleted tasks:

app/code/Magento/CmsGraphQl/etc/schema.graphqls
```diff
@@ -15,7 +15,8 @@ type Query {
         id: Int @doc(description: "Id of the CMS page")
     ): CmsPage @resolver(class: "Magento\\CmsGraphQl\\Model\\Resolver\\Page") @doc(description: "The CMS page query returns information about a CMS page") @cache(cacheTag: "cms_p", cacheIdentity: "Magento\\CmsGraphQl\\Model\\Resolver\\Page\\Identity")
     cmsBlocks (
-        identifiers: [String] @doc(description: "Identifiers of the CMS blocks")
+        identifiers: [String] @doc(description: "Identifiers of the CMS blocks"),
+        scope_code: [String] @doc(description: "Scope Code of the CMS blocks")
```
TomashKhamlai, line 19: Case: send
```graphql
query {
  storeConfig {
    id
    website_id
  }
  cmsBlocks(
    identifiers: null
    scope_code: null
  ) {
    items {
      content
    }
  }
}
```
and get this `"message": "\"identifiers\" of CMS blocks should be specified"`
send another
```graphql
query {
  storeConfig {
    id
    website_id
  }
  cmsBlocks(
    identifiers: ""
    scope_code: null
  ) {
    items {
      content
    }
  }
}
```
and get "message": "Scope Code of CMS blocks should be specified"
then send
```graphql
query {
  storeConfig {
    id
    website_id
  }
  cmsBlocks(
    identifiers: ""
    scope_code: ""
  ) {
    items {
      content
    }
  }
}
```
and get "The CMS block with the \"\" ID doesn't exist."

@naydav, why `identifiers` is not required field. What is the point of sending empty strings. I have seen many times that sending "" or null results in error that some field is not-nullable. Maybe the better logic is to check if the property is omitted and thn address it to `all` or `default` equivalent in REST rather than allowing empty strings?

GovindaSharma:
hi @TomashKhamlai thanks for your review.Just wanted to know only unit test i have to fix right? or is there any other thing i have to check also?

TomashKhamlai:
@GovindaSharma, in GraphQL we cover with api-functional tests, but fixtures are taken from integration. To learn more check accepted/merged PRs or better visit these pages:
[Web API](https://devdocs.magento.com/guides/v2.3/get-started/web-api-functional-testing.html)
[GraphQL](https://devdocs.magento.com/guides/v2.3/graphql/functional-testing.html)

And if you have some free time, I invite you join our conversation #635, because the issue you are trying to fix was reproducible but now it isn't. Perhaps it is still reproducible in certain cases.

TomashKhamlai:
Generally, it works. You tried to implement scope selection as an input parameter, but we already use headers for defining the scope. @GovindaSharma, do you want to continue to work on this issue? I've added a [comment](https://github.com/magento/graphql-ce/issues/635#issuecomment-500864479) today.

Please respond to this message. We are thinking about closing this PR but your feedback is important for us.

In case you are reading this message when PR is closed, don't be shy to create another PR for the same issue. Thank you for your interest.

naydav:
> Generally, it works. You tried to implement scope selection as an input parameter, but we already use headers for defining the scope. @GovindaSharma, do you want to continue to work on this issue? I've added a [comment](https://github.com/magento/graphql-ce/issues/635#issuecomment-500864479) today.
> 
> Please respond to this message. We are thinking about closing this PR but your feedback is important for us.
> 
> In case you are reading this message when PR is closed, don't be shy to create another PR for the same issue. Thank you for your interest.
@GovindaSharma 

Agee with @TomashKhamlai. The scope should be passed as a header.
https://devdocs.magento.com/guides/v2.3/graphql/send-request.html
So we need to change implementation.

PR #737, opened 2019-06-12
486 add customer account validation in quote operations extra

vovsky:
### Description (*)
Verifying the case when we unlock locked customer and perform Quote operations


naydav:
Closed due to all changes were made in https://github.com/magento/graphql-ce/pull/714

PR #745, opened 2019-06-17
Setting visibility to protected for easier extendability

raivisdejus:
### Description (*)
Setting visibility to `protected` to make different graph ql classes easier to extend while creating customer specific adjustments or customizations.

Current visibility of `private` forces us to copy the whole class and that makes custom code unnecessary large and hard to read and update.

If conceptual move from `private` to `protected` is acceptable, I can adjust this PR with changes in other classes as well. 

There are no changes to the functionality

### DIFF
--- a/app/code/Magento/CatalogGraphQl/Model/Resolver/Category/Products.php
+++ b/app/code/Magento/CatalogGraphQl/Model/Resolver/Category/Products.php
@@ -23,17 +23,17 @@ class Products implements ResolverInterface
     /**
      * @var \Magento\Catalog\Api\ProductRepositoryInterface
      */
-    private $productRepository;
+    protected $productRepository;
 
     /**
      * @var Builder
      */
-    private $searchCriteriaBuilder;
+    protected $searchCriteriaBuilder;
 
     /**
      * @var Filter
      */
-    private $filterQuery;
+    protected $filterQuery;
 
     /**
      * @param ProductRepositoryInterface $productRepository
--- a/app/code/Magento/CatalogGraphQl/Model/Resolver/Products.php
+++ b/app/code/Magento/CatalogGraphQl/Model/Resolver/Products.php
@@ -25,22 +25,22 @@ class Products implements ResolverInterface
     /**
      * @var Builder
      */
-    private $searchCriteriaBuilder;
+    protected $searchCriteriaBuilder;
 
     /**
      * @var Search
      */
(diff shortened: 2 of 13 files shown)


vpodorozh:
Hello @raivisdejus, Thank you for contribution!
I see what do you mean, however, the technical vision of Magento has the other direction for solving extensibility issues. Take a look at this guide: https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html
`2.6. Inheritance SHOULD NOT be used. Composition SHOULD be used for code reuse.`
`2.7. All non-public properties and methods SHOULD be private.`

**So in case, you need to extend something - you need to do it in composition way, but not inheritance.**

To improve the extensibility you can split some class to different services - so it will be easier to substitute them using the dependency injection. For example: `Magento\CatalogGraphQl\Model\Resolver\Products\Query\Search:paginateList` can be moved to separate service in case you need to have a possibility to change encapsulated behaviour.

In the end, I can say - sometimes I feel your pain too :) But it is much preferable to use composition to minimize coupling > as result technical debt while the upgrades. (and this is not the only reason)

I'm going to close this PR - re-open it please in case you have the reasons. Thank you! 😊

vpodorozh:
Hello @raivisdejus, Thank you for contribution!
I see what do you mean, however, the technical vision of Magento has the other direction for solving extensibility issues. Take a look at this guide: https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html
`2.6. Inheritance SHOULD NOT be used. Composition SHOULD be used for code reuse.`
`2.7. All non-public properties and methods SHOULD be private.`

**So in case, you need to extend something - you need to do it in composition way, but not inheritance.**

To improve the extensibility you can split some class to different services - so it will be easier to substitute them using the dependency injection. For example: `Magento\CatalogGraphQl\Model\Resolver\Products\Query\Search:paginateList` can be moved to separate service in case you need to have a possibility to change encapsulated behaviour.

In the end, I can say - sometimes I feel your pain too :) But it is much preferable to use composition to minimize coupling > as result technical debt while the upgrades. (and this is not the only reason)

I'm going to close this PR - re-open it please in case you have the reasons. Thank you! 😊

vpodorozh:
By the way, here is a good principle called `decomposition` that describes my suggestion to split class with different services (I just forgot to mention it).

Thx @novikor for reminder.

PR #760, opened 2019-07-02
Correct SendFriend Visibility Check

pmclain:
### Description (*)
The existing check would return an error if sending a product with search only
visibility.

1. Create product with search only visibility
2. Locate product sku using search query
3. Use send a friend mutation to sent search only product

### DIFF
--- a/app/code/Magento/SendFriendGraphQl/Model/Provider/GetVisibleProduct.php
+++ b/app/code/Magento/SendFriendGraphQl/Model/Provider/GetVisibleProduct.php
@@ -49,8 +49,9 @@ class GetVisibleProduct
             $product = $this->productRepository->getById($productId);
 
             if (!in_array(
-                $product->getVisibility(),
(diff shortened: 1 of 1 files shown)


app/code/Magento/SendFriendGraphQl/Model/Provider/GetVisibleProduct.php
```diff
@@ -49,8 +49,9 @@ public function execute(int $productId): ProductInterface
             $product = $this->productRepository->getById($productId);
 
             if (!in_array(
-                $product->getVisibility(),
-                $this->visibility->getVisibleInCatalogIds()
+                (int) $product->getVisibility(),
+                $this->visibility->getVisibleInSiteIds(),
```
pmclain, line 53: The docblock hint for this function's return value is incorrectly `string[]` proposed fix here https://github.com/magento/magento2/pull/23532

naydav (approved):

Issue #892, opened 2019-09-03
[Checkout] Improve strict typing in CartAddressInterface 


paliarush:
The following fields should be marked with `!` in `CartAddressInterface`:
 - `city`
 - `country`
 - `firstname`
 - `lastname`
 - `street`
 - `telephone`

The same fields are required in the input type `CartAddressInput` that's why it is safe to assume that they will not be empty when queried.

Additionally, all fields in `CartAddressRegion` and `CartAddressCountry` must be marked with `!`

Issue #864, opened 2019-08-28
The shipping address contains `null` values when there's no shipping address set on a cart


dplaton:
### Preconditions (*)

1. Magento 2.3.2

### Steps to reproduce (*)

1. Use your favorite GraphQL client to create a guest cart using the `createEmptyCart` mutation
2. Request the cart details (just the shipping address should suffice):
```query {
  cart(cart_id: "YaMkSTdY7fOlShTKVJTNCLZHL2dtAWpT") {
    shipping_addresses {
      city
      company
      country {
        code
      }
      firstname
      lastname
      postcode
      region {
        code
      }
      street
      telephone
    }
  }
}
```

### Expected result (*)

1. The shipping address in the response is an empty array

```
"data": {
    "cart": {
      "shipping_addresses": [ ]
    }
  }
```

### Actual result (*)

1. The shipping address is an object with `null` values

```
  "data": {
    "cart": {
      "shipping_addresses": [
        {
          "city": null,
          "company": null,
          "country": {
            "code": null
          },
          "firstname": null,
          "lastname": null,
          "postcode": null,
          "region": {
            "code": null
          },
          "street": [
            ""
          ],
          "telephone": null
        }
      ]
    }
  }
```

PR #897, opened 2019-09-04
graphQl-892: Improved strict typing in CartAddressInterface

VitaliyBoyko:
### Description (*)
Improved strict typing in CartAddressInterface.

1. #892: [Checkout] Improve strict typing in CartAddressInterface


VitaliyBoyko:
Looks like we have the same issue for the billing address as for the shipping address here #864.

VitaliyBoyko:
Hi @paliarush please clarify what should be returned in the scope of this case \Magento\GraphQl\Quote\Customer\GetSpecifiedBillingAddressTest::testGetSpecifiedBillingAddressIfBillingAddressIsNotSet. So far it returns object with nullable fields. As far as I have understood it should be null instead of an object.

lenaorobei (changes requested):
@VitaliyBoyko billing address should be `null`. Please consider fixing api-functional test.

app/code/Magento/QuoteGraphQl/Model/Resolver/BillingAddress.php
```diff
@@ -44,7 +44,13 @@ public function resolve(Field $field, $context, ResolveInfo $info, array $value
         $cart = $value['model'];
 
         $billingAddress = $cart->getBillingAddress();
-        if (null === $billingAddress) {
+        if (null === $billingAddress ||
```
lenaorobei: Could you please use `validateAddressFromSchema` method? I mean move it to the separate class from  `\Magento\QuoteGraphQl\Model\Resolver\ShippingAddresses` and reuse for both - shipping and billing.

lenaorobei (changes requested):
Please see my suggestion with `validateAddressFromSchema` method.

lenaorobei (approved):

PR #939, opened 2019-09-17
[BUGFIX] Return the url from the customUrlLocator when it is an array

lewisvoncken:
- for example wehn you have a module which has a Router to the routerList it doesn't have url_rewrites but with the urlLocators it is possible to also route these url_keys

### Description (*)

I want to implement the customUrlLocator functionality for Mage2gen. I have taken a look at several blog modules and these even't implemented url_rewrites but a custom router.
https://github.com/krukas/Mage2Gen/pull/305

In the attachment there is an example module which has implemented the customUrlLocator.
[Mage2gen_BlogGraphQl.zip](https://github.com/magento/graphql-ce/files/3623589/Mage2gen_BlogGraphQl.zip)

image

image

image

### DIFF
--- a/app/code/Magento/UrlRewriteGraphQl/Model/Resolver/EntityUrl.php
+++ b/app/code/Magento/UrlRewriteGraphQl/Model/Resolver/EntityUrl.php
@@ -63,6 +63,9 @@ class EntityUrl implements ResolverInterface
             $url = ltrim($url, '/');
         }
         $customUrl = $this->customUrlLocator->locateUrl($url);
+        if (is_array($customUrl)) {
+            return $customUrl;
+        }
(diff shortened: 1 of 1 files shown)


lenaorobei:
Sorry but we cannot accept this PR because Magento API is `\Magento\UrlRewriteGraphQl\Model\Resolver\UrlRewrite\CustomUrlLocatorInterface` and value cannot be `array`since `locateUrl` method returns `string`.
```
public function locateUrl($urlKey): ?string
```

The way how intended behaviour can be archived is to implement new field in schema.

lewisvoncken:
@lenaorobei I will create another pr with a different solution. Thank you for your time!!

Issue #783, opened 2019-07-19
Integrate GraphQL Playground


paliarush:
### Description
To improve dev experience, it should be possible to integrate [GraphQL Playground in-browser IDE](https://github.com/prisma/graphql-playground). 

### Expected behavior
When developer opens `http://magento.instance/graphql` in browser, GraphQL Playground should be opened. It should be possible to execute GraphQL queries in browser without installation of GraphiQL or other GraphQL IDEs.

PR #992, opened 2019-10-08
#783 graphql playground integration

dani97:
Graphql Playground integration with magento 2. 
Accessible in default mode and developer mode.
### Description (*)

1. Graphql Playground integration with magento 2. 
2. Created a route with id graphiql (please suggest a better url, I couldn't think of better one)
3. Accessible in default mode and developer mode as {BASE_URL/graphql} in modern browsers. In production mode, redirection to 404 page.
4. Removed all default magento styles from the page.
5. Implemented as independent module like swagger so can be removed or disabled in production.

1. magento/graphql-ce#783:  Integrate GraphQL Playground

1. In developer or default mode access {BASE_URL}/graphiql
2. In Production mode same url will be redirected to 404 page.

Screenshot from 2019-10-08 20-24-16

 - [X] Pull request has a meaningful description of its purpose
 - [X] All commits are accompanied by meaningful commit messages
 - [x] All new or changed code is covered with unit/integration tests (if applicable)
 - [x] All automated tests passed successfully (all builds are green)


dani97:
No idea how to fix the failing composer.json test.

TomashKhamlai:
Hi, @dani97. May I perform manual functional testing now or it is still work in progress?

dani97:
@TomashKhamlai I have been fixing static tests failure. They are code style issues. I can hold on when you are testing. Thank you.

dani97:
@TomashKhamlai only composer static tests are failing with latest build, you can continue testing.

dani97:
@TomashKhamlai  please update when you are done with manual testing, I will then update my PR for failing static tests. :)

TomashKhamlai:
image
Please, advise something.

I checked out your PR by id and switched to `production mode` in order to verify that schema is loading. Then I run

rm composer.lock && composer install
php bin/magento setup:upgrade
rm -rf var/* generated/* pub/static/*
php bin/magento dep:mo:set developer
php bin/magento s:d:c

dani97:
@TomashKhamlai can you check a js file and css loading from cdn.jsdelivr.net.

TomashKhamlai:
image

TomashKhamlai:
Please try doing this:
```shell
rm -rf var/* pub/static/*
php bin/magento setup:sta:dep -f
```
and then check again.

dani97:
@TomashKhamlai all working great now. can you check again. I will hold fixing the failing bundle tests failure until you are done. Thanks

TomashKhamlai:
Please fix this

```javascript
Uncaught TypeError: require.config is not a function
    at graphiql:66
(anonymous) @ graphiql:66

```

TomashKhamlai:
In general, it works as a standalone application even when I brake the schema. Keep in mind that when you take a task without the label **ready for dev** it can take a lot of time for maintainers to get to your PR. Please, be patient.

dani97:
@TomashKhamlai Thanks I will wait.

dani97:
> Please fix this
> 
> ```js
> Uncaught TypeError: require.config is not a function
>     at graphiql:66
> (anonymous) @ graphiql:66
> ```

I followed the same way where in swagger removed unwanted js modules for the models. The same error is thrown there too on removing requirejs.

TomashKhamlai:
I suggest checking what happens if you keep require-config.js instead of removing). I think that you will have a more effective conversation with code reviewers. Good luck and thank you for your contribution!

lenaorobei:
@paliarush we need you review here as well.

app/code/Magento/GraphqlPlayground/Controller/Index/Index.php
```diff
@@ -0,0 +1,84 @@
+<?php declare(strict_types=1);
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\GraphqlPlayground\Controller\Index;
```
paliarush: For consistency module name should be changed.
```suggestion
namespace Magento\GraphQlPlayground\Controller\Index;
```

app/code/Magento/GraphqlPlayground/view/frontend/templates/playground.phtml
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+// @codingStandardsIgnoreFile
+
+/** @var string $graphqlEndpoint Magento default Graphql Endpoint */
+$graphqlEndpoint =  $block->getBaseUrl() . "graphql";
```
paliarush: `graphql` frontname can be changed and should not be hardcoded here. See https://github.com/magento/magento2/blob/2.3/app/code/Magento/GraphQl/etc/di.xml#L21

app/code/Magento/GraphqlPlayground/etc/frontend/routes.xml
```diff
@@ -0,0 +1,15 @@
...
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:noNamespaceSchemaLocation="urn:magento:framework:App/etc/routes.xsd">
+    <router id="standard">
+        <route id="graphiql" frontName="graphiql">
```
paliarush: We must not introduce a new URL. The idea was to render GraphiQL, when a user opens `/graphql` route in the browser instead of GraphiQL or similar client. With the custom endpoint the value of this feature is pretty low, because developers will have to know about this alternative URL, and will still not getting any help when trying to open `graphql` in the browser.

Please investigate how it is possible to avoid new endpoint.

paliarush (changes requested):

app/code/Magento/GraphqlPlayground/etc/frontend/routes.xml
dani97: @paliarush  can we use **user-agent** header to differentiate, I can think of that as the only possibility. Can you suggest if there is a better approach, or any native magento utility to achieve this.

app/code/Magento/GraphqlPlayground/view/frontend/web/css/playground.min.css
```diff
@@ -0,0 +1 @@
+body{margin:0;padding:0;font-family:sans-serif;overflow:hidden}#root{height:100%}body{font-family:Open Sans,sans-serif;-webkit-font-smoothing:antialiased;-moz-osx-font-smoothing:grayscale;color:rgba(0,0,0,.8);line-height:1.5;height:100vh;letter-spacing:.53px;margin-right:-1px!important}a,body,code,h1,h2,h3,h4,html,p,pre,ul{margin:0;padding:0;color:inherit}a:active,a:focus,button:focus,input:focus{outline:none}button,input,submit{border:none}button,input,pre{font-family:Open Sans,sans-serif}code{font-family:Consolas,monospace}
```
lenaorobei: Both versions - minified and not minified should be added.

lenaorobei: Same for `js`.

app/code/Magento/GraphqlPlayground/view/frontend/templates/playground.phtml
```diff
@@ -0,0 +1,20 @@
...
+?>
+<div id="root" class="playgroundIn">
+</div>
+<script>
+     const root = document.getElementById('root');
```
lenaorobei: ES6 not supported in Magento, please do not use `const`.

app/code/Magento/GraphqlPlayground/view/frontend/web/css/playground.min.css
dani97: @lenaorobei  shall I add the map file here.

app/code/Magento/GraphqlPlayground/etc/frontend/routes.xml
DrewML: > can we use user-agent header to differentiate, I can think of that as the only possibility.

Would prefer we avoid doing anything with user agent strings - notoriously unreliable.

Can we use the following requirements for a request:

- Is a `GET`
- Does not have a `query` querystring value

If both are true, render the playground

app/code/Magento/GraphqlPlayground/view/frontend/web/css/playground.min.css
DrewML: I'd actually prefer that we do not check in _any_ compiled code in this module

This is a dev-only feature, so minification isn't really important here - would prefer to opt for clearer exceptions for bug reports.

app/code/Magento/GraphqlPlayground/etc/frontend/routes.xml
dani97: @DrewML Thanks for the suggestions. This is more clear than user-agent.

app/code/Magento/GraphqlPlayground/view/frontend/web/css/playground.min.css
dani97: @DrewML I took reference from `https://github.com/prisma-labs/graphql-playground
https://github.com/prisma-labs/graphql-playground/blob/master/packages/graphql-playground-html/minimal.html` I have used this as a reference. I can add a react component using this `https://github.com/prisma-labs/graphql-playground#as-react-component`. Is this fine?

DrewML: @dani97 the way you're setting it up now is fine - just swap in the unminified version of the CSS/JS

lenaorobei:
@dani97 could you please pull latest `2.3-develop` and address review comments?

dani97:
> @dani97 could you please pull latest `2.3-develop` and address review comments?

@lenaorobei  I have fixed most of the comments.I could not find the unminified version from graphql playground, have to dig more to find how they transpiled and minified the react components, so I will add map file.

TomashKhamlai:
@dani97, one little note. When built-in GraphQL Playground opens, I see the store header present in the location bar of the built-in client.
image
I ask to change it. The expected behavior is that the header is set in the "HEADERS" section at the bottom of the page and nowhere else. More about accepting GraphQL headers in Magento can be read in the [Documentation: Send Request #request-headers](https://devdocs.magento.com/guides/v2.3/graphql/send-request.html#request-headers)

dani97:
> @dani97, one little note. When built-in GraphQL Playground opens, I see the store header present in the location bar of the built-in client.
> image
> I ask to change it. The expected behavior is that the header is set in the "HEADERS" section at the bottom of the page and nowhere else. More about accepting GraphQL headers in Magento can be read in the [Documentation: Send Request #request-headers](https://devdocs.magento.com/guides/v2.3/graphql/send-request.html#request-headers)

Thanks for pointing out. Fixed it @TomashKhamlai

lenaorobei:
@magento run all tests

chalov-anton:
Each time I try to **Execute** (Ctrl+Enter)an empty query, a "_Syntax Error GraphQL request (2:1) Unexpected <EOF>\n\n1: # Write your query or mutation here\n2: \n   ^\n_" appears in Wed Browser Console 
empty query_graphql_console

dani97:
@chalov-anton  It is not a issue I believe, also graphql playground has this same behavior out of the box.

chalov-anton:
Considering the PR Description https://github.com/magento/graphql-ce/pull/992#issue-325814054 the Graphql Playground integration should not be accessible from **Production mode**
But now it is
prod_mode_integration
(The browser cache has been cleaned)

lenaorobei:
As per conversation with @dani97 we agreed to open a new PR to https://github.com/magento/magento2 repo.

@dani97 thanks for understanding and collaboration!

